PR #22, opened 2020-06-17
Rename some minor parts of the ECDSA spec

robdockins:
to avoid name clashes with the upcoming Cryptol release.

### DIFF
--- a/Common/EC/ec_point_ops.cry
+++ b/Common/EC/ec_point_ops.cry
@@ -13,7 +13,7 @@ nzAffinePoint (ax,ay) = { x = ax, y = ay }
 type JacobianPoint fv = { x : fv, y : fv, z : fv }
 
 /* Create affine point from Jacobian (only well-defined if z coordinate is non-zero). */
-ec_affinify : {fv} (Field fv, JacobianPoint fv) -> AffinePoint fv
+ec_affinify : {fv} (FieldRep fv, JacobianPoint fv) -> AffinePoint fv
 ec_affinify(f,s) = { x = f.mul(g2, s.x), y = f.mul(g3, s.y) }
   where g = f.div(f.field_unit, s.z)
         g2 = f.sq(g)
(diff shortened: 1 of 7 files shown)


robdockins:
Since I opened this PR, I discovered we have a second, parallel ECDSA spec that is based explicitly on `Z p` types.  It also requires some fixing to work with the new split-arith Cryptol.  Most critically, the changes to the exponentiation operator now cause defaulting to occur for the concrete exponent values in expressions like:

```
mp_mod_sqrt : Z p -> Z p
mp_mod_sqrt c = r'' ^^ (2 ^^ 94)
  where t1  = (c ^^ 2) * c
        t2  = (t1 ^^ (2 ^^ 2)) * t1
        t3  = (t2 ^^ (2 ^^ 4)) * t2
        t4  = (t3 ^^ (2 ^^ 8)) * t3
        r   = (t4 ^^ (2 ^^ 16)) * t4
        r'  = (r ^^ (2 ^^ 32)) * c
        r'' = (r' ^^ (2 ^^ 96)) * c
```

Because of the current defaulting rules, expressions like `2^^96` get defaulted to type `[2]` which is... not what we want.  As a result the modular square root algorithms are all messed up.

atomb:
Oh, yeah, that's definitely not what we want! Do you have any thoughts about alternative defaulting that would work better? In cases with constants, it seems that the size of the base is very likely to be much smaller than the desired size of the result in general. I haven't thought through things very thoroughly yet, but I'm kind of leaning in the direction of decimal values not defaulting to bit vectors at all.

robdockins:
Yeah, I think the basic rule that makes sense is: never default to a limited-precision type.  The new defaulting rules that @yav and I have been talking about all have this property, I think.  So, the exponents here would be defaulted to `Integer`, which then behaves as expected.

atomb:
That sounds great.

PR #27, opened 2020-09-16
Port SHAImp specs to parameterized SHA.cry specs

bboston7:
This PR removes `SHA256Imp.cry` and `SHA512Imp.cry` and ports their functionality to `SHA.cry`.  Users of instantiations of the `SHA.cry` module (`SHA256.cry`, `SHA512.cry`, etc) should be unaffected by these changes.  However, users of `SHA.cry` will likely experience breakage if they have written their own model instantiations.  The potentially breaking changes are:

* Introduction of a new parameter `digest_bytes` that must be provided.  This will break complete model instantiations outside of this repo.
* Stricter type constraints on `w` and `digest_size`.  However, these new restrictions are minor and I don't see them causing breakage in most cases.


Primitive/Keyless/Hash/SHA.cry
```diff
@@ -18,11 +18,19 @@ parameter
...
 
   type digest_size : #
   type constraint (fin digest_size, 8*w >= digest_size)
 
+  // `digest_bytes` exists to skirt around a Cryptol bug that causes loss of
+  // information in parametrized modules (Cryptol GitHub issue #883).  Using
+  // `digest_size/8` in place of `digest_bytes` or defining
+  // `type digest_bytes = digest_size/8` will trigger the bug and cause an
+  // unsolved constraint error in `HMAC.cry`.
+  type digest_bytes : #
```
andreistefanescu, line 31: now this is unnecessary

andreistefanescu (approved):
👍

Primitive/Keyless/Hash/SHA.cry
bboston7: Good catch, fixed.

PR #44, opened 2021-12-14
skip nightly CI for now

pnwamk:
While things are being worked on having
nightly reminders that things are still broken is
unhelpful noise. When we get to the CI passing,
nightly builds can be enabled again to help us
spot if changes to Cryptol have broken
things here.


RyanGlScott (approved):
Here's to a clearer inbox.

Issue #46, opened 2022-03-21
`Primitive::Symmetric::Cipher::Authenticated::GCM` incorrect


abakst:
The following test, if added to `Primitive::Symmetric::Cipher::Authenticated::AES_GCM`, fails:

```cryptol
property enc_dec_inverse_4 = dec.valid && (pt == dec.pt)
   where
     key = 0xeebc1f57487f51921c0465665f8ae6d1658bb26de6f8a069a3520293a572078f
     iv  = 0x99aa3e68ed8173a0eed06684
     pt  = 0xf56e87055bc32d0eeb31b2eacc2bf2a5
     aad = 0x4d23c3cec334b49bdb370c437fec78de
     enc = gcmEnc `{K=256, IV=96, AAD=128, T=128} {E=encrypt} {key=key, iv=iv, aad=aad, pt=pt}
     dec = gcmDec `{K=256, IV=96, AAD=128, T=128} {E=encrypt} {key=key, iv=iv, aad=aad, ct=enc.ct, tag=enc.tag}
```

The issue is that `enc_dec` in the `GCM` module is identical in the encryption and decryption case, but the spec referenced by the module (<https://luca-giuzzi.unibs.it/corsi/Support/papers-cryptography/gcm-spec.pdf>) indicates that the tag computation differs between encrypt/decrypt:
> The authenticated decryption operation is similar to the encrypt operation, but with the order of
the hash step and encrypt step reversed.

The fix should be straightforward, which is to add a bit to `enc_dec` to select between the different encrypt/decrypt modes

jldodds:
I think I agree that it's wrong but it was a little hard to understand why. It's because we use C and P and then just try to swap them with the arguments of ```gcmEnc``` and ```gcmDec``` but of course that doesn't change them inside the function and you end up passing the plaintext into the tag for decryption, since that ends up in the ```C``` variable. I propose that instead of trying to be so clever we just make two functions like the paper does so you can see what's what.

for reference--
The encrypt functions from the paper
![image](https://user-images.githubusercontent.com/667803/160471348-f30294a4-1790-4d65-9c72-a65e3688ccd0.png)
The decrypt functions from the paper
![image](https://user-images.githubusercontent.com/667803/160471447-f318fc80-dfea-4715-8e2f-096df8bcc636.png)

weaversa:
In case you're looking for one that follows the NIST spec, I'll leave this here.

[...]

PR #47, opened 2022-03-21
Fix error in `GCM` decryption

abakst:
Address #46


marsella:
This was superseded by the changes in #75, which solve the same problem in a different way.

PR #49, opened 2023-02-09
Fixes to make things work with the new module system

yav:


andrew-bivin:
@yav Are there any more bottlenecks to getting this merged? My interest is in having CI function / also because our PQ specs work with the new module system.

yav:
The new module system is not yer merged in `master`, so if we were to merge this, we'd have to somehow modify the CI to run against the `functors-merge` branch of Cryptol

andrew-bivin:
> The new module system is not yer merged in `master`, so if we were to merge this, we'd have to somehow modify the CI to run against the `functors-merge` branch of Cryptol

Got it. Do you have a rough idea of when `functors-merge` will be ready? I'm trying to get an idea of how/when this will all fit together. I know funding for these things is spread across multiple projects/PDI so it's just not clear to me

PR #52, opened 2023-05-03
Add `wots_PKfromSK` function to SPHINCS+ spec

bboston7:
This change adds a function to derive a WOTS public key from a secret key to the SPHINCS+ spec. It also adds a property asserting that this new function is equivalent to deriving a public key from the secret key seed.

The motivation for this addition is that it is faster than `wots_PKgen` when the secret has already been generated.


mariosge (approved):
Looks good to me and the property passes!

PR #54, opened 2023-05-11
Parallelize `chain` function calls in SPHINCS+ spec

bboston7:
This change replaces sequence comprehensions involving the `chain` function with equivalent `parmap` expressions. The resulting WOTS implementation is significantly faster when run with multithreading enabled. To enable multithreading, invoke `cryptol` with the following runtime system option:

```
$ cryptol +RTS -N<threads> -RTS
```

I tested this change by running `:check wotsCorrectness` and `:check wotsPKCheck`, both of which pass.


andrew-bivin:
@bboston7 Do you still need review on this before merging? Or are there additional commits planned

bboston7:
Ideally yes, a review would be nice

bboston7:
No additional commits are planned. This is ready to go

eddywestbrook (approved):
All of the changes look straightforward, and assuming that Brett tested that the new versions of the specs, this should be fine to merge. Thanks Brett!

PR #58, opened 2023-07-12
Fix Kyber KEM

### LINKED ISSUE #55: Kyber and known answer tests
I tested the Kyber Cryptol spec for decryption using the known answer tests from NIST.  To do it, I created an AWK script that reads the NIST responses and generates a Cryptol batch file.  Submitting that file to Cryptol shows that for every one of the 100 known answer tests, the answers disagree.

There are three attachments.  The first one is the AWK script (you have to remove the .txt extension).  Second one is a bash script that sets the CRYPTOLPATH (you have to remove the .txt extension and make it executable). The final file is a transcript that shows how to create the batch file and the results of running it.

[kat.awk.txt](https://github.com/GaloisInc/cryptol-specs/files/11602017/kat.awk.txt)
[cryptol-specs.txt](https://github.com/GaloisInc/cryptol-specs/files/11602028/cryptol-specs.txt)
[kyber.txt](https://github.com/GaloisInc/cryptol-specs/files/11602052/kyber.txt)

mariosge:
This PR changes the behavior of Kyber's KEM.Enc and KEM.Dec to output each key byte in reverse order.
As a result:
- The specs now pass the known-answer tests
- The KEM Correctness property passes on random inputs

The solution is not ideal because it slightly deviates from the actual specs but it solves the issue for now. Perhaps a better solution is to modify `BitsToBytes` to account for the reverse ordering.

Resolves #55

### DIFF
--- a/Primitive/Asymmetric/Cipher/Kyber/3.01/specification.tex
+++ b/Primitive/Asymmetric/Cipher/Kyber/3.01/specification.tex
@@ -777,7 +777,7 @@ we define key generation, encapsulation, and decapsulation of \KyberCCAKEM.
 \begin{code}
   // We make the random message m explicit
   CEnc : ([12*k*n/8+32]Byte, [32]Byte) -> ([d_u*k*n/8+d_v*n/8]Byte, [inf]Byte)
-  CEnc (pk, m) = (c,K) where
+  CEnc (pk, m) = (c, map reverse K) where
     m' = H(m) // Cryptol warns about shadowing
     (KBar,r) = G(m'#H(pk))
     c = Enc(pk,m',r)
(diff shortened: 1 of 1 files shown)


Primitive/Asymmetric/Cipher/Kyber/3.01/specification.tex
```diff
@@ -777,7 +777,7 @@ \subsection{Specification of \KyberCCAKEM}
 \begin{code}
   // We make the random message m explicit
   CEnc : ([12*k*n/8+32]Byte, [32]Byte) -> ([d_u*k*n/8+d_v*n/8]Byte, [inf]Byte)
-  CEnc (pk, m) = (c,K) where
+  CEnc (pk, m) = (c, map reverse K) where
```
bboston7, line 780: Because this deviates from the spec, can you please add a comment explaining why the `map reverse` is necessary?

Primitive/Asymmetric/Cipher/Kyber/3.01/specification.tex
```diff
@@ -809,8 +809,7 @@ \subsection{Specification of \KyberCCAKEM}
 
 \begin{code}
   CDec : ([d_u*k*n/8+d_v*n/8]Byte,[24*k*n/8+96]Byte) -> [inf]Byte
-  CDec (c,sk) = if (c == c') then KDF(KBar'#H(c))
-                             else KDF(z#H(c))
+  CDec (c,sk) = map reverse K
```
bboston7, line 812: Similarly, a comment explaining the `map reverse` would be nice here.

bboston7 (approved):

Primitive/Asymmetric/Cipher/Kyber/3.01/specification.tex
mariosge: @bboston7 I just added the related comments. If this looks OK to you, should I merge?

PR #60, opened 2023-08-01
Kyber bytes to bits

mariosge:
This draft PR edits the Cryptol function `BytesToBits` so that the bits are in reverse order and consistent with the original specs. Currently all tests pass but the KAT fail.


marsella:
I'm going to close this PR; it is now outdated with the update to ML-KEM's final version and the removal of the older version in #217.

PR #67, opened 2023-12-18
Literate Cryptol specifications

carrala:
This PR replaces some Cryptol implementations of algorithms with literate Cryptol versions which include the text from the original specifications as well as any tests that were defined for the algorithms.

Updated algorithms include:
- DES
- LED
- PRINCE
- 3DES
- Salsa20
- ZUC

All tests for the new implementations pass locally.


Primitive/Symmetric/Cipher/Stream/ZUC1_6.md
```diff
@@ -0,0 +1,423 @@
...
+                        , ((s@4)  <<< 20)
+                        , ((s@0)  <<< 8)
+                        , (s@0) ]
+```
+
+and the same implementation is needed for step 1 of the function LFSRWithWorkMode.
+
+
+*Informative note:* For two elements a, b over GF(2^31-1), the computation of `v=a+b mod (2^31-1)` can be done by
+
```
RyanGlScott, line 169: Is this newline after "can be done by" intentional? I don't think this is present in the original spec linked above, and it looks somewhat strange in the rendered Markdown.

Primitive/Symmetric/Cipher/Stream/Salsa20.md
```diff
@@ -0,0 +1,449 @@
...
+
+### Inputs and outputs
+
+Let k be a 32-byte or 16-byte sequence. Let v be an 8-byte sequence. Let m be an l-byte sequence for some `l ∈ {0,1,...,2^70}`. The **Salsa20 encryption of** m **with nonce** v **under key** k, denoted Salsa20k(v) ⊕ m, is an l-byte sequence.
+
+Normally k is a secret key (preferably 32 bytes); v is a nonce, i.e., a unique message number; m is a plaintext message; and Salsa20k(v) ⊕ m is a ciphertext message. Or m can be a ciphertext message, in which case Salsa20k(v) ⊕ m is the original plaintext message.
+
+```cryptol
+Salsa20Encrypt : {a, l} (a >= 1, 2 >= a, l >= 0, 2^^70 >= l)
+       => [16*a][8] -> [8][8] -> [l][8] -> [l][8]
```
RyanGlScott, line 416: I noticed that this PR changes the API, as `Salsa20_encrypt` has been renamed to `Salsa20Encrypt`. (There might be other API changes as well—I have not looked deeply.) Is this intentional? I'm somewhat worried about introducing backwards-incompatible name changes without a good reason.

Primitive/Symmetric/Cipher/Stream/Salsa20.md
```diff
@@ -0,0 +1,449 @@
...
+        k0 = [1..16]
+        k1 = [201..216]
+        n  = [101..116]
+```
+
+### Comments
+
+"Expansion" refers to the expansion of (k,n) into Salsa20k(n). It also refers to the expansion of k into a long stream of Salsa20k outputs for various n's; see Section 10.
+
+The constants `σ0σ1σ2σ3` and `τ0τ1τ2τ3` are "expand 32-byte k" and "expand 16-byte k" in ASCII.
```
RyanGlScott, line 404: This is using inline code formatting for `σ0σ1σ2σ3` and `τ0τ1τ2τ3`, but neither of those are valid Cryptol expressions. Perhaps they should be `σ0 # σ1 # σ2 # σ3` and `τ0 # τ1 # τ2 # τ3` instead?

RyanGlScott (commented):
Thanks, @carrala! This is a nice improvement. I haven't read everything line by line (mainly because most of the text is taken directly from upstream specifications), but I did take a closer look at the Salsa20 implementation, which is the one that I'm most familiar with.

Primitive/Symmetric/Cipher/Stream/Salsa20.md
RyanGlScott: For what it's worth, I wasn't totally sure about `cryptol-specs`' stance on backwards-incompatible changes myself, so I asked @yav (who is more experienced with `cryptol-specs` than I am). @yav's stance is that we shouldn't be afraid to make API changes right now if it would lead to cleaner specs going forward.

That being said, we are already departing from the spec a bit by choosing the more descriptive name "`Salsa20_encrypt`" rather than simply "`Salsa20`" (the latter being what the spec uses). In that case, we may as well stick with the existing `Salsa20_encrypt` name, which this PR now does after your most recent commit.

RyanGlScott (commented):
Thanks, @carrala! This looks good on my end. Can you resolve the merge conflicts as well?

carrala:
@RyanGlScott All merge conflicts have been resolved!

RyanGlScott (approved):

PR #69, opened 2023-12-26
Add Implementation of ML-KEM

mariosge:
This PR implements the NIST standard ML-KEM as specified in https://csrc.nist.gov/pubs/fips/203/ipd

The parameterized implementation is in the `specification.cry` file. Details regarding the spec appear in the README file.
The specs pass all KATs as they appear in the official NIST files as well as 3rd party KATs.

The PR also updates the global README file to point to the correct directory.


eddywestbrook (approved):
This LGTM. I'm not an expert in this algorithm, but because it passes the KATs and also matches the on-paper FIPS spec I feel pretty confident in it. Also, having the links to the references and the PDF for the spec here are especially helpful. Thanks Marios!

PR #70, opened 2024-01-19
Add check and prove steps to CI

carrala:
- Check stage: finds and attempts to check all Cryptol properties. It fails if any of the properties fail.
- Prove stage: finds and attempts to prove all Cryptol properties with a time limit. It records all properties that don't complete in the time period in artifact `manual_proves.txt`; all successful proves are saved in `fast_proves.txt`. It fails if any of the properties that completed in the time period fail.
- A handful of failing or hanging properties with Cryptol files are updated to pass these checks.


RyanGlScott:
I haven't taken a close look at this yet, as it seems like there are some CI-related issues to sort out. Feel free to ping me when you'd like a closer review, however.

marsella:
This has been superseded by a more limited CI that checks docstrings on some specified files. We'll probably build off that infrastructure if we want to, say, make a nightly job that tries checking properties on all the specs.

PR #72, opened 2024-02-27
Merge `sha-imperative` branch

RyanGlScott:
This brings in the changes from the long-lived [`sha-imperative`](https://github.com/GaloisInc/cryptol-specs/tree/sha-imperative) branch, but rebased on top of the current state of the `master` branch.


eddywestbrook (approved):
All this does is add a number of new algorithms and change the key sizes for the AES and DRBG algorithms to use larger keys, so it should be fine to merge it in

PR #75, opened 2024-06-12
aes-gcm: fix decryption bug, add properties #46

### LINKED ISSUE #46: `Primitive::Symmetric::Cipher::Authenticated::GCM` incorrect

### LINKED ISSUE #74: Fix bug in AES-GCM spec
We received a bug report for AES-GCM:
> We’ve noticed that the `enc_dec` function is assumed to be symmetric. Obviously for the AES-CTR part this holds true, but for the GHASH it does not. [Noting here](https://github.com/GaloisInc/cryptol-specs/blob/master/Primitive/Symmetric/Cipher/Authenticated/GCM.cry#L66) .
> 
> The tag is produced from the output data. When performing an encrypt, then the output is the cipher text and so the output is correct. But for a decrypt the ghash would be on the plain text, and so is incorrect.

We should already have the standard set of test vectors encoded, but they must not have caught this case. We should try to identify a test case that does trigger a failure and add it to the repo.

Related: #46 also suggests a bug in AES-GCM's `enc_dec` function. I haven't thought carefully enough to decide whether it's the same bug.

## To do:
- [ ] Fix this bug
- [ ] Determine whether this fix also addresses #46
- [ ] Add a test case to confirm correctness
- [ ] Decide whether to augment the AES-GCM spec into a Literate version; if so, make another issue.

marsella:
Closes #46. Closes #74. Closes #47.

Here's an update to the AES-GCM spec that fixes the decryption bug (decryption was just implemented incorrectly, and it was never tested / proven to work anyway). 
The existing version of AES is based on what I think is a submission document for the GCM algorithm. The format, algorithm names, and parameter names were changed at some point, and the official NIST standard uses some slightly different notation. I kept the old APIs in because it seems like they're being used, but changed the references and everything else to adhere to the NIST spec.

Conveniently, there were several correct implementations floating around:
- In the original issue, folks proposed and implemented (#47) adding a flag to this version to make decryption correct
- Also in the original issue, [someone provided a different implementation](https://github.com/GaloisInc/cryptol-specs/issues/46#issuecomment-1085018600) that more closely adheres to the AES-GCM NIST spec. This forms the basis of most of the changes in this PR.

This PR:
- fixes the spec to be correct
- update the properties to validate decryption (to some degree, see following questions)
- adds a NIST-adherent API and states an equivalence property to the existing API
- adds a basic encryption / decryption is symmetric property
- adds a batch file to prove the test vectors and check the new and existing properties. I have not actually proven any of the properties that are currently `:check`ed in the file; I added docs with a note of how long I waited.

This is now [ready for review!](https://github.com/GaloisInc/cryptol-specs/pull/75#pullrequestreview-2123614190)
---
These questions have been answered:
- The version from `weaversa` had a requirement that P, IV, and A are all finite. Is that redundant if we have size upper bounds on each of them? Does it still allow negative infinity? Is the fact that it's used as a length enforce that it's non-negative?
- Style question: Taking / returning a named record vs a set of parameters. Why would we do one vs the other?
- How can we make two type variables and constraint them to be the same, so that things look more like the human spec? (P = len_C)
- Guidelines on when to use module-level parameters vs function-level?
- How do I go about doing something useful with the equivalence properties in gcm.cry? Should I prove these in the batch file for some typical parameter sets? Is there a way to say for all?

I want to look into the following items:
- I think the set of acceptable tag lengths is informed by the other parameters, or there might be acceptable sets in general and it should disallow other kinds (beyond just the individual length requirements). See if this is true and if we can enforce it
- GHASH. Check if 128 is correct fixed size for hash or if that should be a parameter as well
- I think there are additional properties that we could demo here as well, like wrong tag => invalid decryption (statistically at least)

### DIFF
--- a/Primitive/Symmetric/Cipher/Authenticated/AES_GCM.cry
+++ b/Primitive/Symmetric/Cipher/Authenticated/AES_GCM.cry
@@ -18,9 +18,7 @@ property GCM_AE_test_vector_0 = ct == expected_ct /\ tag == expected_tag /\ vali
         dec = GCM_AD `{K=128, IV=96, AAD=0, T=128} {E=AES::encrypt} zero zero ct [] tag
         expected_ct = []
         expected_tag = 0x58e2fccefa7e3061367f1d57a4e7455a
-        valid_dec = case dec of
-            Some actual_pt -> pt == actual_pt
-            None -> False
+        valid_dec = dec.valid && (dec.pt == pt)
 
 property GCM_AE_test_vector_1 =
     ct == expected_ct /\ tag == expected_tag /\ valid_dec
@@ -33,9 +31,7 @@ property GCM_AE_test_vector_1 =
         expected_tag = 0xab6e47d42cec13bdf53a67b21257bddf : [128]
         (ct, tag) = GCM_AE `{K=128, IV=96} {E=AES::encrypt} key iv pt aad
         dec = GCM_AD `{K=128, IV=96} {E=AES::encrypt} key iv ct aad tag
-        valid_dec = case dec of
-            Some actual_pt -> pt == actual_pt
-            None -> False
+        valid_dec = dec.valid && (dec.pt == pt)
 
 property GCM_AE_test_vector_2 =
     ct == expected_ct /\ tag == expected_tag /\ valid_dec
@@ -48,9 +44,7 @@ property GCM_AE_test_vector_2 =
         expected_tag = 0x4d5c2af327cd64a62cf35abd2ba6fab4 : [128]
         (ct, tag) = GCM_AE {E=AES::encrypt} key iv pt aad
         dec = GCM_AD {E=AES::encrypt} key iv ct aad tag
-        valid_dec = case dec of
-            Some actual_pt -> pt == actual_pt
-            None -> False
+        valid_dec = dec.valid && (dec.pt == pt)
 
 property GCM_AE_test_vector_3 =
     ct == expected_ct /\ tag == expected_tag /\ valid_dec
@@ -63,9 +57,7 @@ property GCM_AE_test_vector_3 =
         expected_tag = 0x5bc94fbc3221a5db94fae95ae7121a47
         (ct, tag) = GCM_AE {E=AES::encrypt} key iv pt aad
         dec = GCM_AD {E=AES::encrypt} key iv ct aad tag
-        valid_dec = case dec of
-            Some actual_pt -> pt == actual_pt
-            None -> False
+        valid_dec = dec.valid && (dec.pt == pt)
 
 // A test case from aesgcmtest.c
 property GCM_AE_test_vector_4 =
@@ -79,12 +71,10 @@ property GCM_AE_test_vector_4 =
         expected_tag  = 0x67ba0510262ae487d737ee6298f77e0c
         (ct, tag) = GCM_AE {E=AES::encrypt} key iv pt aad
         dec = GCM_AD {E=AES::encrypt} key iv ct aad tag
-        valid_dec = case dec of
-            Some actual_pt -> pt == actual_pt
-            None -> False
+        valid_dec = dec.valid && (dec.pt == pt)
 
 property GCM_AE_invalid_test_vector =
-    ct == expected_ct /\ tag == expected_tag /\ ~valid_dec
+    ct == expected_ct /\ tag == expected_tag /\ ~dec.valid
     where
         key   = 0xeebc1f57487f51921c0465665f8ae6d1658bb26de6f8a069a3520293a572078f : [256]
         iv    = 0x99aa3e68ed8173a0eed06684 : [96]
@@ -95,6 +85,3 @@ property GCM_AE_invalid_test_vector =
         invalid_tag  = 0x67ba0510262ae487d737ee6298f77888
         (ct, tag) = GCM_AE {E=AES::encrypt} key iv pt aad
         dec = GCM_AD {E=AES::encrypt} key iv ct aad invalid_tag
-        valid_dec = case dec of
-            Some actual_pt -> pt == actual_pt
-            None -> False
--- a/Primitive/Symmetric/Cipher/Authenticated/AES_GCM.cry
+++ b/Primitive/Symmetric/Cipher/Authenticated/AES_GCM.cry
@@ -1,37 +1,100 @@
 
 // Cryptol AES GCM test vectors
-// Copyright (c) 2010-2018, Galois Inc.
+// Copyright (c) 2010-2024, Galois Inc.
 // www.cryptol.net
 // Author: Ajay Kumar Eeralla
 
-//Test vectors from http://luca-giuzzi.unibs.it/corsi/Support/papers-cryptography/gcm-spec.pdf
+// Test vectors from http://luca-giuzzi.unibs.it/corsi/Support/papers-cryptography/gcm-spec.pdf
 
 module Primitive::Symmetric::Cipher::Authenticated::AES_GCM where
 import `Primitive::Symmetric::Cipher::Authenticated::GCM
-import Primitive::Symmetric::Cipher::Block::AES_parameterized
+import Primitive::Symmetric::Cipher::Block::AES_parameterized as AES
 
 
-property testPass0 = gcmEnc `{K=128, IV=96, AAD=0, T=128} {E=encrypt} {key=zero, iv=zero, pt=[], aad=[]} ==
-                              {ct = [], tag = 0x58e2fccefa7e3061367f1d57a4e7455a}
+property GCM_AE_test_vector_0 = ct == expected_ct /\ tag == expected_tag /\ valid_dec
+    where
+        pt = []
+        (ct, tag) = GCM_AE `{K=128, IV=96, AAD=0, T=128} {E=AES::encrypt} zero zero pt []
+        dec = GCM_AD `{K=128, IV=96, AAD=0, T=128} {E=AES::encrypt} zero zero ct [] tag
+        expected_ct = []
+        expected_tag = 0x58e2fccefa7e3061367f1d57a4e7455a
+        valid_dec = case dec of
+            Some actual_pt -> pt == actual_pt
+            None -> False
 
-property testPass1 = gcmEnc `{K=128, IV=96, AAD=0, T=128} {E=encrypt} {key=zero, iv=zero, pt=zero, aad=[]} ==
-                              {ct = 0x0388dace60b6a392f328c2b971b2fe78, tag = 0xab6e47d42cec13bdf53a67b21257bddf}
+property GCM_AE_test_vector_1 =
+    ct == expected_ct /\ tag == expected_tag /\ valid_dec
+    where
+        key = zero
+        iv = zero
+        pt = zero
+        aad = []
+        expected_ct = 0x0388dace60b6a392f328c2b971b2fe78
+        expected_tag = 0xab6e47d42cec13bdf53a67b21257bddf : [128]
+        (ct, tag) = GCM_AE `{K=128, IV=96} {E=AES::encrypt} key iv pt aad
+        dec = GCM_AD `{K=128, IV=96} {E=AES::encrypt} key iv ct aad tag
+        valid_dec = case dec of
+            Some actual_pt -> pt == actual_pt
+            None -> False
 
-property testPass2 = gcmEnc `{K=128, IV=96, AAD=0, T=128} {E=encrypt} {key=0xfeffe9928665731c6d6a8f9467308308, iv=0xcafebabefacedbaddecaf888,         pt=0xd9313225f88406e5a55909c5aff5269a86a7a9531534f7da2e4c303d8a318a721c3c0c95956809532fcf0e2449a6b525b16aedf5aa0de657ba637b391aafd255, aad=[]} ==
-    {ct = 0x42831ec2217774244b7221b784d0d49ce3aa212f2c02a4e035c17e2329aca12e21d514b25466931c7d8f6a5aac84aa051ba30b396a0aac973d58e091473f5985,
-    tag = 0x4d5c2af327cd64a62cf35abd2ba6fab4}
+property GCM_AE_test_vector_2 =
+    ct == expected_ct /\ tag == expected_tag /\ valid_dec
+    where
+        key = 0xfeffe9928665731c6d6a8f9467308308 : [128]
+        iv = 0xcafebabefacedbaddecaf888 : [96]
+        pt = 0xd9313225f88406e5a55909c5aff5269a86a7a9531534f7da2e4c303d8a318a721c3c0c95956809532fcf0e2449a6b525b16aedf5aa0de657ba637b391aafd255
+        aad = []
+        expected_ct = 0x42831ec2217774244b7221b784d0d49ce3aa212f2c02a4e035c17e2329aca12e21d514b25466931c7d8f6a5aac84aa051ba30b396a0aac973d58e091473f5985
+        expected_tag = 0x4d5c2af327cd64a62cf35abd2ba6fab4 : [128]
+        (ct, tag) = GCM_AE {E=AES::encrypt} key iv pt aad
+        dec = GCM_AD {E=AES::encrypt} key iv ct aad tag
+        valid_dec = case dec of
+            Some actual_pt -> pt == actual_pt
+            None -> False
 
-property testPass3 = gcmEnc `{K=128, IV=96, AAD=160, T=128} {E=encrypt} {key=0xfeffe9928665731c6d6a8f9467308308,                                            iv=0xcafebabefacedbaddecaf888,
-    pt=0xd9313225f88406e5a55909c5aff5269a86a7a9531534f7da2e4c303d8a318a721c3c0c95956809532fcf0e2449a6b525b16aedf5aa0de657ba637b39,
-    aad=0xfeedfacedeadbeeffeedfacedeadbeefabaddad2} ==
-    {ct=0x42831ec2217774244b7221b784d0d49ce3aa212f2c02a4e035c17e2329aca12e21d514b25466931c7d8f6a5aac84aa051ba30b396a0aac973d58e091,
-    tag=0x5bc94fbc3221a5db94fae95ae7121a47}
+property GCM_AE_test_vector_3 =
+    ct == expected_ct /\ tag == expected_tag /\ valid_dec
+    where
+        key = 0xfeffe9928665731c6d6a8f9467308308
+        pt  = 0xd9313225f88406e5a55909c5aff5269a86a7a9531534f7da2e4c303d8a318a721c3c0c95956809532fcf0e2449a6b525b16aedf5aa0de657ba637b39
+        iv  = 0xcafebabefacedbaddecaf888
+        aad = 0xfeedfacedeadbeeffeedfacedeadbeefabaddad2
+        expected_ct  = 0x42831ec2217774244b7221b784d0d49ce3aa212f2c02a4e035c17e2329aca12e21d514b25466931c7d8f6a5aac84aa051ba30b396a0aac973d58e091
+        expected_tag = 0x5bc94fbc3221a5db94fae95ae7121a47
+        (ct, tag) = GCM_AE {E=AES::encrypt} key iv pt aad
+        dec = GCM_AD {E=AES::encrypt} key iv ct aad tag
+        valid_dec = case dec of
+            Some actual_pt -> pt == actual_pt
+            None -> False
 
 // A test case from aesgcmtest.c
-// Test passes
-property testPass4 = gcmEnc `{K=256, IV=96, AAD=128, T=128} {E=encrypt} {key=0xeebc1f57487f51921c0465665f8ae6d1658bb26de6f8a069a3520293a572078f,
-    iv=0x99aa3e68ed8173a0eed06684, pt=0xf56e87055bc32d0eeb31b2eacc2bf2a5, aad=0x4d23c3cec334b49bdb370c437fec78de} ==
-   {ct=0xf7264413a84c0e7cd536867eb9f21736, tag=0x67ba0510262ae487d737ee6298f77e0c}
+property GCM_AE_test_vector_4 =
+    ct == expected_ct /\ tag == expected_tag /\ valid_dec
+    where
+        key   = 0xeebc1f57487f51921c0465665f8ae6d1658bb26de6f8a069a3520293a572078f : [256]
+        iv    = 0x99aa3e68ed8173a0eed06684 : [96]
+        pt    = 0xf56e87055bc32d0eeb31b2eacc2bf2a5 : [128]
+        aad   = 0x4d23c3cec334b49bdb370c437fec78de : [128]
+        expected_ct   = 0xf7264413a84c0e7cd536867eb9f21736
+        expected_tag  = 0x67ba0510262ae487d737ee6298f77e0c
+        (ct, tag) = GCM_AE {E=AES::encrypt} key iv pt aad
+        dec = GCM_AD {E=AES::encrypt} key iv ct aad tag
+        valid_dec = case dec of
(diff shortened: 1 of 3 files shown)


RyanGlScott (commented):
> The version from `weaversa` had a requirement that P, IV, and A are all finite. Is that redundant if we have size upper bounds on each of them?

I assume you mean bounds like `P <= 2^^39 - 256`? If so, then yes, that implies `fin P`, so it would be redundant to say `(fin P, P <= 2^^39 - 256)`.

> Does it still allow negative infinity? Is the fact that it's used as a length enforce that it's non-negative?

All type-level numbers are non-negative, so you don't need to worry about negative infinity.

> Style question: Taking / returning a named record vs a set of parameters. Why would we do one vs the other?

This is ultimately a matter of taste, so there aren't any hard guidelines. That being said, I think that if you start using tuples that are sufficiently large (say, four or more elements), then I would generally prefer using records at that point. Naming the fields makes it easier to tell at a glance what each field represents, and it confers the advantage that you can construct the record without needing to remember the exact order in which the fields were defined.

> How can we make two type variables and constraint them to be the same, so that things look more like the human spec? (P = len_C)

I'm not quite sure what you mean here. Can you give an example of what you are trying to do?

> Guidelines on when to use module-level parameters vs function-level?

In general, you'd make something a module-level parameter if it is shared among multiple functions in the module (e.g., the key size `K`). Again, I think this comes down to a matter of taste.

> How do I go about doing something useful with the equivalence properties in gcm.cry? Should I prove these in the batch file for some typical parameter sets? Is there a way to say for all?

You could try `:check`'ing some of the properties. For certain properties, you might even be able to cover all possible inputs using `:prove`, but I imagine that some properties will simply be too complicated to `:prove` in a reasonable amount of time (this is where you'd need SAW to do better). But even just `:check` would be better than nothing.

Primitive/Symmetric/Cipher/Authenticated/GCM.cry
```diff
@@ -29,82 +30,182 @@ parameter
...
-enc_dec :
-  {n} ((2^^39 - 256) >= n) => [K] -> [IV] -> [AAD] -> [n] -> ([n], [T])
-enc_dec K IV A P = (C,T)
+ * GCM-AD Function, [NISTSP800-38D] Section 7.2, Algorithm 5. This provides
+ * authenticated decryption.
+ */
+
+GCM_AD : { C }
+    ( fin C, C <= 2^^39 - 256 )  // This constraint comes from [NISTSP800-38D]
+ => [K] -> [IV] -> [C] -> [AAD] -> [T] -> Option [C]
```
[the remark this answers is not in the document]
marsella: Oh darn. That's good to know. The spec says to return the plaintext or FAIL; is the canonical way to do this without enums to use the valid / pt record that gcmDec returns?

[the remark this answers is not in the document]
RyanGlScott: Something like that, yes.

Primitive/Symmetric/Cipher/Authenticated/GCM.cry
```diff
@@ -29,82 +30,182 @@ parameter
...
+    // Set names to match the spec
+    CIPHk = E k
+    len_IV = `IV
+    len_A = `AAD
+    len_P = `P
+
+    H = CIPHk 0
+    type s = IV %^ 128  // Equivalently: 128 * IV /^ 128 - IV
+    J0 = if len_IV == 96 then coerceSize iv # (0 : [31]) # (1 : [1])
+         else GHASH`{IV /^ 128 + 1} H (iv # (0 : [s + 64]) # (len_IV: [64]))
```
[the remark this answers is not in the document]
marsella: This is much easier to understand, thanks for writing it out for me!

marsella:
> > How can we make two type variables and constraint them to be the same, so that things look more like the human spec? (P = len_C)
> I'm not quite sure what you mean here. Can you give an example of what you are trying to do?

I was missing the type keyword. The method uses parameter `P` to define the length of the plaintext and ciphertext, while the NIST spec uses `len(C)`. I was trying to figure out how to add a variable called `len_C` to use in different operations that had the same value as `P`, but I was trying to do it value-wise (`len_C ='P`) when I should've been defining a type variable (`type len_C = P`).

Primitive/Symmetric/Cipher/Authenticated/GCM.cry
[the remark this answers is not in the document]
marsella: Thank you! This looks much nicer.

marsella:
Here are some follow-up notes:

- The parameter choices in the implementation are a bit looser than what's described in the implementation. Specifically:
	- P17: `t` may be any one of the following five values: 128, 120, 112, 104, or 96
	- P15: the block size shall be 128 bits
	- P15: the key size shall be at least 128 bits
	- P16: len(P) $\le 2^{39}-256$; len(P) is a multiple of 8
	- P16: len(A) $\le 2^{64}-1$; len(A) is a multiple of 8
	- P16: $1 \le \text{len(IV)} \le 2^64-1$; len(IV) is a multiple of 8
	- encrypt and decrypt have to support the same len(C, A, IV)s.
- Parameter architecture: I looked at AES_parameterized but it looks to be more flexible than the actual set of NIST-approved parameters ([FIPS 197, update 1](https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.197-upd1.pdf)). The expected parameters are encoded in AES.cry so I don't understand what the parameterized version is doing. ValidKey constraint seems too flexible; doesn't it have to be exactly the expected key length?
	- Maybe this is what Joe was mentioning about parameterizations; currently you have to manually change the line in AES.cry to change between parameter sets, but it would be nicer to just define the 3 sets and be done.
- The GHASH signature is correct.
- Properties: there are some additional properties we could prove here. Initial thoughts:
	- Do we write negative tests? I added one property that asserts that the wrong tag will fail, but that's not super compelling.
	- We could try proving symmetry `dec (enc x) = x`. If we did so, I'd like to pick the most-common parameter sets used in the wild (plaintext and aad length are kind of a wildcard here but key / IV / tag length likely have only a few common configs used in practice).
	- I don't think this GCM is used for any other block ciphers, but if it was we could test them
	- Lots of tests are commented out in AES.cry
	- GCTR: if X is empty, return Y = empty string. This is obvious from observation but we could still test it
		- related: do we write properties for private functions? Do those just live outside the private block even though nobody would use them outside the batch file?
	- Cryptol is not suited to proving properties over time, right? As in, the IV is a nonce and must be chosen uniquely across different encryptions, but there's no way to specify that in the AE function.

RyanGlScott:
I've attempted to answer the questions that I feel qualified to answer:

> Do we write negative tests? I added one property that asserts that the wrong tag will fail, but that's not super compelling.

I am generally a fan of more tests, although I don't have a suggestion for how to make it more compelling than what you currently have.

> We could try proving symmetry dec (enc x) = x.

Yes, I agree that that is a pretty compelling property to test, even if we only test it on specific inputs.

> Lots of tests are commented out in AES.cry

I can provide at least a partial explanation for this, since I recently modified the code in `AES.cry`. In https://github.com/GaloisInc/cryptol-specs/commit/0a60f79bdc007f4cd59a573ecb1fbb577b54abd5, the key size for AES was changed from 128 bits to 256 bits. Before that change, only the 128-bit tests were used, and tests written with other key sizes in mind were commented out. To keep the existing convention, that commit comments out the 128-bit tests and enables the 256-bit tests.

We could, in principle, enable the tests at other key sizes as well, but I think we'd need to parameterize the code in `AES.cry` more in order to make this possible, as everything in `AES.cry` currently assumes 256-bit key sizes. (Whether this level of parameterization is desired or not is unclear to me.)

> GCTR: if X is empty, return Y = empty string. This is obvious from observation but we could still test it 

Yes, we may as well test this.

> related: do we write properties for private functions? Do those just live outside the private block even though nobody would use them outside the batch file?

We can certainly test properties over private functions. Note that Cryptol doesn't yet have a clean way to mark properties as private (see https://github.com/GaloisInc/cryptol/issues/1621), but that needn't be a blocker.

> Cryptol is not suited to proving properties over time, right? As in, the IV is a nonce and must be chosen uniquely across different encryptions, but there's no way to specify that in the AE function.

I don't think there is any direct way to prove that, other than perhaps checking if the particular values of `IV` at specific instantiations of AES-GCM are distinct.

marsella:
Thank you! Appreciate the context on the AES parameterization.

I think there's a more general question there about whether we want to support smaller AES key sizes and if so, whether we're satisfied with the given structure -- 128 is mostly not recommended in practice but it's still part of Suite B so we might want to have easier access to it. I was looking at the Kyber implementation which is generic over parameters and then has a few modules that just define the standardized parameter sets. This is definitely out of scope for this PR but we might consider it if we come at this piece in the future.



> > Cryptol is not suited to proving properties over time, right? As in, the IV is a nonce and must be chosen uniquely across different encryptions, but there's no way to specify that in the AE function.
>
> I don't think there is any direct way to prove that, other than perhaps checking if the particular values of IV at specific instantiations of AES-GCM are distinct.

I was wondering more if there's a way to say, like, a type is a 128-bit string *drawn uniformly at random* but that seems very hard to mechanically reason about.

Primitive/Symmetric/Cipher/Authenticated/aes_gcm.bat
```diff
@@ -0,0 +1,26 @@
...
+:prove AES_GCM_test_vector_4
+:prove AES_GCM_invalid_test_vector
+
+// The following checks do not really provide any significant formal
+// verification because they check so little of the sample space.
+
+// These properties can be checked manually; one of the APIs calls the other.
+// They take more than an hour to prove.
+:check decryptionApisAreEquivalent `{K=128, IV=96, AAD=0, T=128, P=256} {E=AES::encrypt}
+:check encryptionApisAreEquivalent `{K=128, IV=96, AAD=0, T=128, P=256} {E=AES::encrypt}
```
marsella, line 16: style question: Should I parameterize these in the AES-GCM module with AES so we don't have to do it manually here? I did that for `aesGcmIsSymmetric`

Primitive/Symmetric/Cipher/Authenticated/aes_gcm.bat
```diff
@@ -0,0 +1,26 @@
...
+
+// This property is independent of the type parameters but we have to specify
+// them anyway.
+// It takes more than 25 minutes to prove.
+:check dotAndMultAreEquivalent `{K=128, IV=96, AAD=0, T=128} {E=AES::encrypt}
```
marsella, line 21: note: None of these parameters are relevant because this is just a function on 128-bit blocks. Should I make a separate batch file for it? Is there a way to reorganize the GCM module so that this property and the two relevant functions are not bound by the module parameters?

Primitive/Symmetric/Cipher/Authenticated/GCM.cry
```diff
@@ -29,82 +30,204 @@ parameter
...
+  */
+  define_J0 : { len_IV } ( len_IV == IV) => [K] -> [len_IV] -> [128] -> [128]
+  define_J0 k iv H
+    | len_IV == 96 => iv # (0 : [31]) # (1 : [1])
+    | len_IV != 96 => GHASH`{len_IV /^ 128 + 1} H (iv # (0 : [s + 64]) # (`len_IV: [64]))
+    where
+      // Set names to match the spec
+      type s = len_IV %^ 128  // Equivalently: 128 * len_IV /^ 128 - len_IV
+
+  /**
```
marsella, line 163: Note: everything from here down, as well as most of the `GCM_AE` and `GCM_AD` methods are pulled from the implementation @weaversa provided in #46. Do we have a standard way of citing this? Does cryptol-specs try to maintain the authorship notes at the top of the file?

Primitive/Symmetric/Cipher/Authenticated/AES_GCM.cry
```diff
@@ -1,37 +1,89 @@
...
+        iv    = 0x99aa3e68ed8173a0eed06684 : [96]
+        pt    = 0xf56e87055bc32d0eeb31b2eacc2bf2a5 : [128]
+        aad   = 0x4d23c3cec334b49bdb370c437fec78de : [128]
+        expected_ct   = 0xf7264413a84c0e7cd536867eb9f21736
+        expected_tag  = 0x67ba0510262ae487d737ee6298f77e0c
+        (ct, tag) = GCM_AE {E=AES::encrypt} key iv pt aad
+        dec = GCM_AD {E=AES::encrypt} key iv ct aad tag
+        valid_dec = dec.valid && (dec.pt == pt)
 
+property AES_GCM_invalid_test_vector =
```
marsella, line 78: note: This is just a single negative test made by borking the tag from the fourth test vector. It's not exactly superseded by the symmetric property because that doesn't guarantee that there's only one correct tag for a given `(key, iv, ct)` tuple. I'm going to make a note of this in the subsequent clean-up issue.

marsella (commented):
This is now ready for review!

I welcome nit-picky feedback on naming, spacing, and style from those who are more familiar with Cryptol standards than I am (e.g. everyone here).

Before I merge this I am going to make a separate issue collecting the concerns I noted in a previous comment so they don't get lost.

Primitive/Symmetric/Cipher/Authenticated/GCM.cry
```diff
@@ -29,82 +30,204 @@ parameter
...
+/**
+ * Property demonstrating equivalence between `gcmEnc` and `GCM_AE`.
+ */
+encryptionApisAreEquivalent : {P} ( fin P, P <= 2^^39 - 256 )
+  => [K] -> [IV] -> [P] -> [AAD] -> Bool
+property encryptionApisAreEquivalent key iv pt aad = gcm_ae_output == (gcmEnc_output.ct, gcmEnc_output.tag)
+    where
+      input = { key=key, iv=iv, pt=pt, aad=aad }
+      gcmEnc_output = gcmEnc input
+      gcm_ae_output = GCM_AE key iv pt aad
```
marsella, line 133: Here are the definitions of the equivalence properties. They're parameterized and called in the batch file.

I expected these to be fast to prove because `gcmEnc` calls `GCM_AE`, but they are not.

Primitive/Symmetric/Cipher/Authenticated/GCM.cry
```diff
@@ -29,82 +30,204 @@ parameter
...
+ * Property demonstrating that decryption is the inverse of encryption.
+ *
+ * This should be instantiated for each block cipher `E` that uses GCM mode.
+ */
+gcmIsSymmetric : {P} ( fin P, P < 2^^39 - 256 )
+  => [K] -> [IV] -> [P] -> [AAD] -> Bool
+property gcmIsSymmetric key iv pt aad = dec.valid && (dec.pt == pt)
+    where
+        (ct, tag) = GCM_AE key iv pt aad
+        dec = GCM_AD key iv ct aad tag
```
marsella, line 145: Here's the symmetric property. It's parameterized in `AES_GCM.cry` and called in the batch file. I expected this to be fast-ish to prove because most of the steps of `GCM-AE` are duplicated in `GCM-AD` but it is not.

marsella (commented):
@mccleeary-galois We talked a few days ago about proving properties that I thought would be fast but were not actually fast in practice. Not sure how much of a priority it is to dive into the SMT statements for these, but I'll tag them below for ya.

Primitive/Symmetric/Cipher/Authenticated/GCM.cry
RyanGlScott: I think it would be nice to credit Sean in the copyright info above.

Primitive/Symmetric/Cipher/Authenticated/aes_gcm.bat
```diff
@@ -0,0 +1,26 @@
...
+// verification because they check so little of the sample space.
+
+// These properties can be checked manually; one of the APIs calls the other.
+// They take more than an hour to prove.
+:check decryptionApisAreEquivalent `{K=128, IV=96, AAD=0, T=128, P=256} {E=AES::encrypt}
+:check encryptionApisAreEquivalent `{K=128, IV=96, AAD=0, T=128, P=256} {E=AES::encrypt}
+
+// This property is independent of the type parameters but we have to specify
+// them anyway.
+// It takes more than 25 minutes to prove.
```
RyanGlScott, line 20: Do you have a sense for what is causing it to be so slow?

In any case, it may be wise to reduce the number of random tests that we do in Cryptol to make this complete in a more reasonable amount of time. You can do this in a Cryptol batch file by running:

```
:set tests=<N>
```

The default value of `<N>` is 100, so we could certainly stand to lower it if these tests take an excessively long time to complete.

Primitive/Symmetric/Cipher/Authenticated/aes_gcm.bat
RyanGlScott: We could imagine splitting up the module hierarchy more to make `dotAndMultAreEquivalent` only parameterized over `K`. That being said, it's unclear that doing so would pay its weight. Personally, I think it's fine to just invoke `dotAndMultAreEquivalent `{K=128, IV=96, AAD=0, T=128}` like you do here.

Primitive/Symmetric/Cipher/Authenticated/aes_gcm.bat
RyanGlScott: Sure, that seems reasonable.

RyanGlScott:
> Should I try to prove the 'check`ed properties before merging this?

You can certainly try, but I wouldn't invest too much time in doing so. We've had to spend a considerable amount of effort to prove [similar properties about AES-GCM](https://github.com/GaloisInc/aws-lc-verification) in SAW, and this required manually rewriting proof goals. I would be very surprised if Cryptol could prove properties of a similar level of complexity without any assistance from a proof engineer.

Primitive/Symmetric/Cipher/Authenticated/aes_gcm.bat
marsella: Oh, this isn't clear. The `:check` here is fairly fast, but calling it with `:prove` instead is very slow. I think if we want to claim that this is formally verified we'd need to eventually `:prove` it. I don't know why `:prove` is so slow, but I have no intuition for how Cryptol works or how to form properties that might prove faster or slower.

RyanGlScott: My apologies, I interpreted the comment above as applying to `:check` rather than `:prove`. In that case, the call to `:check` here is fine as-is. (I am not surprised that it takes so long to prove, for reasons mentioned in https://github.com/GaloisInc/cryptol-specs/pull/75#issuecomment-2180850669.)

Primitive/Symmetric/Cipher/Authenticated/GCM.cry
mccleeary-galois: Agreed on this comment.

mccleeary-galois:
> > Should I try to prove the 'check`ed properties before merging this?
> 
> You can certainly try, but I wouldn't invest too much time in doing so. We've had to spend a considerable amount of effort to prove [similar properties about AES-GCM](https://github.com/GaloisInc/aws-lc-verification) in SAW, and this required manually rewriting proof goals. I would be very surprised if Cryptol could prove properties of a similar level of complexity without any assistance from a proof engineer.

Agreed with @RyanGlScott here, I took a peak at this a bit and it would almost certainly require some proof engineering... I think leaving things as :check is good enough for now.

Primitive/Symmetric/Cipher/Authenticated/GCM.cry
mccleeary-galois: How long are these proofs taking at the moment?

Primitive/Symmetric/Cipher/Authenticated/GCM.cry
mccleeary-galois: I think we should settle for a :check in the batch file at the moment.

mccleeary-galois:
This is a great step in the right direction for this! Generally LGTM for what the scope of this ticket is. 
Most of my comments are around future concerns of CI automation.

mccleeary-galois (approved):

Primitive/Symmetric/Cipher/Authenticated/GCM.cry
marsella: I have not run them to completion, but more than 25 minutes. I can set them spinning on my laptop and see if they end by the end of the day.

marsella:
I made #77 and #78 to collect some of the other concerns raised here.

Primitive/Symmetric/Cipher/Authenticated/GCM.cry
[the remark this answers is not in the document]
kiniry: > I am somewhat cautious about using Cryptol enums here, if only because SAW does not yet support them. I bring this up since there is [at least one SAW development](https://github.com/GaloisInc/aws-lc-verification) in the wild that uses this AES-GCM spec, and I am a bit reluctant to introduce a change that would make it incompatible with SAW for the time being.

AFAIK, we are no longer obligated to care about AWS use of SAW in the wild.  They are welcome to continue to use the AES-GCM spec that they have, and past/current versions of SAW, but it is on them to decide if they are going to track our updates to `cryptol-specs` or use future releases of SAW.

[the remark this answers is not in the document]
kiniry: Thus, we *must* begin to use enums in accordance with MM discussions on this topic, particularly wrt coordinated feature additions across clients.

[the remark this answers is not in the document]
marsella: Thanks for the heads up. I added switching back to options to #78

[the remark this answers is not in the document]
RyanGlScott: > AFAIK, we are no longer obligated to care about AWS use of SAW in the wild.

To be clear, I'm not worried about this use case in particular. My concern is more that SAW [does not support Cryptol enums at all](https://github.com/GaloisInc/saw-script/issues/2052) at the moment, so by changing the GCM spec to use Cryptol enums, we are effectively barring ourselves from using the new GCM spec in _any_ SAW developments. That being said, I don't believe we're using the GCM spec anywhere outside of `aws-lc-verification` at the moment, so in this particular instance, this needn't be a blocker. Nevertheless, we should carefully consider where we use Cryptol enums in `cryptol-specs` going forward.

Having said all of this, it would be nice if SAW actually did support Cryptol enums, as that would make this discussion moot. (It would be wonderful if someone could fix https://github.com/GaloisInc/saw-script/issues/2052, as I doubt I'll have time to do so with my current obligations.)

Issue #78, opened 2024-06-21
Update AES-GCM to meet "gold standard" requirements


marsella:
The GCM spec and AES instantiation of it needs a bit of work to meet our "gold standard" for specs. This was initially flagged in discussion on #75; see that PR for more details.

Some things to review:
- [x] ~There's a duplicated version (`AES_256_GCM.cry` and `AES_GCM.cry`). Determine whether we can remove this duplication or improve documentation to explain why we need two versions (maybe blocked on #77)~ Based on discussion, this will now be handled in #86
    - [x] Consider rewriting `GCM.cry` in terms of `Cipher` so that the instantiation is simpler and it can be reused for other (non-AES) block ciphers
    - [x] Consider making separate modules for different key sizes (`AES256_GCM` and `AES128_GCM`)
- [x] Review parameter choices to make sure they are as tight as the NIST spec requires
- [x] Check for properties defined in the NIST spec and make sure they are explicit in the cryptol spec
    - [x] GCTR: if `X` is empty, the return value should be the empty string
- [x] ~Convert into a literate spec~ This will be handled in #87
- [x] Add warnings in the spec and readme about failure modes that Cryptol cannot prevent (IV reuse)
- [x] Switch to using options instead of record return
- [x] Note any additional proposed changes in this issue

marsella:
Here are my notes on the parameter requirements in [the NIST spec](https://nvlpubs.nist.gov/nistpubs/legacy/sp/nistspecialpublication800-38d.pdf):
- P17: `t` may be any one of the following five values: 128, 120, 112, 104, or 96
- P15: the block size shall be 128 bits
- P15: the key size shall be at least 128 bits
- P16: len(P) $\le 2^{39}-256$; len(P) is a multiple of 8
- P16: len(A) $\le 2^{64}-1$; len(A) is a multiple of 8
- P16: $1 \le \text{len(IV)} \le 2^64-1$; len(IV) is a multiple of 8
- encrypt and decrypt have to support the same len(C, A, IV)s.

PR #79, opened 2024-06-24
Simplify AES module structure

marsella:
This addresses some duplication in the AES specifications. Previously, there was a version that was concretely parameterized to be AES256 and a version with a top-level type constraint that could be parameterized to any of AES-128, -192, or -256 (and maybe others as well, I am not totally clear whether it was explicit about the upper bound). This PR does the following:
- Converts the parameterizable version to a functor called `AES_specification.cry` to match a precedent set elsewhere in the repo (see Kyber). It was formerly `AES_parameterized.cry`
- Adds three concrete instantiations of the functor version for the three allowable key sizes, and adds a minimal test vector for each (`AES128.cry`, `AES192.cry`, and `AES256.cry`)
- Replaces all uses of the concretely parameterized version (`AES.cry`) with the AES256 instantiation of the functor, and removes the (now duplicate) concretely parameterized version.
	- One or two specs used `AES.cry` but expected a 128-bit key, so I put in the correct version there.
- Updated the implementations of the modes of operation. Previously, these were just some test vectors that implicitly instantiated AES128. I added explicit functions to each AES-instantiated mode of operation and made it clearer what key size they're using.

This does not preserve any kind of backwards compatibility. @RyanGlScott suggested a migration guide to using the new APIs, which I have not produced yet. My main blocker is that I'm not sure how to write modules that operate over AES (of any key size). E.g. I'd want to write them with respect to `AES_specification` but require you to pick a key size before you can use them.

I welcome feedback across the board: naming and style conventions, architecture review, type constraint design, cryptographic correctness. I am least concerned about the latter, since this does not introduce any new cryptography, just rearranges what was already there. The diff is fairly large, but I tried to keep my commits coherent if you would prefer to review them one by one.

I noted several possible follow-up issues inline. Another few are:
- The modes of operation are written in terms of functions (the block cipher type signature). These should probably be updated to be functors[^func] parameterized over the nice `Cipher` type, which defines the encrypt / decrypt functions. This would hopefully enforce the reality of modes of operation, which is that they are meaningless without an appropriate block cipher to bring them to life.
- There are two AES-GCM implementations. These can hopefully be deduplicated, perhaps in #78.
- The modes of operation instantiated with AES (of all sizes) are missing many test vectors that exist in their specifications.

[^func]: I have acquired one hammer and every file is a nail now.


Primitive/Symmetric/Cipher/Block/AES_specification.cry
```diff
@@ -0,0 +1,56 @@
...
+import `Primitive::Symmetric::Cipher::Block::AES::Algorithm as AES
+import `Primitive::Symmetric::Cipher::Block::AES::ExpandKey
+import Primitive::Symmetric::Cipher::Block::AES::TBox
+import Primitive::Symmetric::Cipher::Block::Cipher(Cipher)
+
+parameter
+   type m : #
+   type constraint (fin m, m <= 2)
+
+type AesKeySize = 64 * (m + 2)
```
marsella, line 22: question: I was trying to come up with a way to set the key size as the parameter instead of `m` (which we still need in order to set a bunch of other parameters) but couldn't get the inference to do what I wanted. The reason for doing so would be so that the parameterized modules are more obvious -- it's not clear why "m=1" means AES-192. The argument for not bothering is that there are only 3 instances and I put documentation in there already.

Primitive/Symmetric/Cipher/Block/AES128.cry
marsella, line 1: question: Is there a nice way to put the three instantiations of AES into one file instead of having three tiny ones? Do we have a style preference on that?

Primitive/Symmetric/Cipher/Block/AES_specification.cry
```diff
@@ -0,0 +1,56 @@
...
+
+   This provides the basic AES block cipher, abstracted over the key length.
+
+   It operates over 128-bit blocks. To use AES with any practical application,
+   use it with a mode of operation, like CTR or GCM-SIV.
+*/
+
+module Primitive::Symmetric::Cipher::Block::AES_specification where
+
+import `Primitive::Symmetric::Cipher::Block::AES::Algorithm as AES
```
marsella, line 13: note: I actually don't understand why this compiles at all because this import further restricts the `Mode` parameter to be greater than 2, and I thought that the `m` in this file is used to instantiate that `Mode`. So I expected that AES-128 and -192 just wouldn't work. But it does 🤷 

In Algorithm.cry, we have
```haskell

parameter
  /** 0: AES128, 1: AES192, 2: AES256 */
  type Mode : #

  type constraint (2 >= Mode)
```

Primitive/Symmetric/Cipher/Authenticated/AES_GCM_SIV.cry
marsella, line 1: note: I constrained AES-GCM-SIV to just use AES256, which is effectively what it was doing before (due to the type constraint below). I think a nicer option would be to write GCM-SIV in terms of a `Cipher` parameter and instantiate it with AES128 and 256.

I also don't think this works right now; I tried dropping a test vector in but it didn't work in either the before or after version and I chose not to spend much time debugging right now. There are some weird constraints here (for example, the RFC allows the associated data to be 0-64 bytes long; here it's restricted to a minimum of 44 bytes?). I think all of this is better suited to a separate issue, which I'm happy to write up. This work does not make anything worse, at least.

Primitive/Symmetric/Cipher/Block/AESKeyWrap.cry
```diff
@@ -8,9 +8,15 @@
 // shifting and the other using indexing.  This specification closely matches
 // the index-based algorithm description as that description is more similar
 // to the BoringSSL AES Key Wrap implementation.
+//
+// One important deviation: this is currently instantiated only for AES256;
+// the RFC supports all three standardized AES key sizes.
```
marsella, line 13: note: this and the padded version should both be generalized over any AES key size. I am not exactly sure of the right architecture to do that -- would we write this as a functor with the same parameter as in `AES_parameterized`? Or like a `Cipher`? Or maybe something else.

In any case, this currently maintains the status quo of only instantiating it for AES256. I can write a separate issue to parameterize this to match the spec if we think it's worth having that recorded.

Primitive/Symmetric/Cipher/Authenticated/AES_GCM.cry
```diff
@@ -8,43 +8,52 @@
...
 import `Primitive::Symmetric::Cipher::Authenticated::GCM
-import Primitive::Symmetric::Cipher::Block::AES_parameterized as AES
+import Primitive::Symmetric::Cipher::Block::AES128 as AES128
+import Primitive::Symmetric::Cipher::Block::AES256 as AES256
+
+AES128_GCM_encrypt = GCM_AE `{K=128} {E=AES128::encrypt}
+AES128_GCM_decrypt = GCM_AD `{K=128} {E=AES128::encrypt}
+
+AES256_GCM_encrypt = GCM_AE `{K=256} {E=AES256::encrypt}
+AES256_GCM_decrypt = GCM_AD `{K=256} {E=AES256::encrypt}
```
marsella, line 18: question: I am trying to partially initialize the functor here -- set the `K` and `E` parameters and leave the others (`T`, `AAD`, `IV`) up to the caller. This compiled the way I expected, but below on L40 & L41, I wasn't able to specify those other parameters when I called it; I had to give explicit types to the parameters themselves. E.g. I thought I could do 

```
AES128_GCM_encrypt `{ IV: [96] } zero zero [] []
```
but I couldn't get the bracket notation to compile. Instead, I had to specify the type of the IV where I declared it:
```
iv = zero: [96]
AES128_GCM_encrypt zero iv [] []
```

Why is that the case? And does this still do what I want it to do (instantiate AES-GCM for a fixed key size and arbitrary IV, plaintext, and tag length)?

Primitive/Symmetric/Cipher/Block/AES_specification.cry
marsella, line 1: note: This is mostly not new; it's almost identical to the old [`AES_parameterized.cry`](https://github.com/GaloisInc/cryptol-specs/pull/79/files#diff-c5f37e0b30eaf94e732bdd878ab6692fc4b70e45af2342d48f8a8b42fcb08ec9) except I added a parameter `m` and changed the definitions of `EncKey` and `DecKey`.

Primitive/Symmetric/Cipher/Block/Modes/AES128_CBC.cry
```diff
@@ -0,0 +1,53 @@
...
+
+module Primitive::Symmetric::Cipher::Block::Modes::AES128_CBC where
+import Primitive::Symmetric::Cipher::Block::Modes::CBC
+import Primitive::Symmetric::Cipher::Block::AES128 as AES128
+
+AES128_CBC_encrypt: {n} (fin n) => [AES128::AesKeySize] -> iv -> [n]block -> [n]block
+AES128_CBC_encrypt = cbcEnc AES128::encrypt
+
+AES128_CBC_decrypt: {n} (fin n) => [AES128::AesKeySize] -> iv -> [n]block -> [n]block
+AES128_CBC_decrypt = cbcDec AES128::decrypt
```
marsella, line 14: note: this is the only thing I added to this file -- concrete names for "cbc mode using AES128." The test vectors are reformatted but the same content as in the old `AES_CBC.cry`.

Primitive/Symmetric/Cipher/Block/Modes/AES128_CFB.cry
```diff
@@ -0,0 +1,37 @@
...
+
+module Primitive::Symmetric::Cipher::Block::Modes::AES128_CFB where
+import Primitive::Symmetric::Cipher::Block::Modes::CFB
+import Primitive::Symmetric::Cipher::Block::AES128 as AES128
+
+AES128_CFB_encrypt: {n} (fin n) => [AES128::AesKeySize] -> iv -> [n]block -> [n]block
+AES128_CFB_encrypt = cfbEnc AES128::encrypt
+
+AES128_CFB_decrypt: {n} (fin n) => [AES128::AesKeySize] -> iv -> [n]block -> [n]block
+AES128_CFB_decrypt = cfbDec AES128::decrypt
```
marsella, line 15: note: ditto here - this is the only new thing in this file compared to `AES_CFB.cry`.

Primitive/Symmetric/Cipher/Block/Modes/AES128_CTR.cry
```diff
@@ -0,0 +1,23 @@
...
+
+module Primitive::Symmetric::Cipher::Block::Modes::AES128_CTR where
+import Primitive::Symmetric::Cipher::Block::Modes::CTR
+import Primitive::Symmetric::Cipher::Block::AES128 as AES128
+
+AES128_CTR_encrypt: {n} (fin n) => [AES128::AesKeySize] -> ic -> [n]block -> [n]block
+AES128_CTR_encrypt = ctrEnc AES128::encrypt
+
+AES128_CTR_decrypt: {n} (fin n) => [AES128::AesKeySize] -> ic -> [n]block -> [n]block
+AES128_CTR_decrypt = ctrDec AES128::decrypt
```
marsella, line 15: note: ditto here - this is the only thing new in this file compared to `AES_CTR.cry`.

Primitive/Symmetric/Cipher/Block/Tests/TestAES.cry
```diff
@@ -0,0 +1,47 @@
+/**
+ * Test vectors for AES256
+ * I don't actually know where these are sourced from.
```
marsella, line 3: note: These test vectors previously lived in `AES.cry`; the 128- and 192-bit ones were commented out.

Primitive/Symmetric/Cipher/Block/AES_specification.cry
weaversa: When I (most recently) specified this I used:

```
/**
 * This constraint enforces the standard key sizes of 128, 192, and
 * 256-bits [FIPS-PUB-197 Sections 1, 5, and 6.1].
 */
type constraint Key k = (fin k, k % 64 == 0, k / 64 >= 2, k / 64 <= 4)
```

marsella:
I added #80 with some of the comments on block ciphers.

mccleeary-galois (approved):
Overall looks like a good step in the right direction! Thanks for doing this @marsella. I think we should try and get the migration guide in and ensure we have the follow on tickets logged but happy with approving this.

Primitive/Symmetric/Cipher/Block/AES_specification.cry
marsella: Oh gosh you're right. I stared at this so many times!

Primitive/Symmetric/Cipher/Block/AESKeyWrap.cry
marsella: Made #81!

Primitive/Symmetric/Cipher/Authenticated/AES_GCM_SIV.cry
marsella: On second look, I misread the comparison in L22. This previously supported both AES versions, but now it only supports one. I'll take another pass and try to fix it.

marsella: Fixed the supporting-multiple-keys issue in [6430913](https://github.com/GaloisInc/cryptol-specs/pull/79/commits/6430913116d947c4edf55432642497309eed5ab5).

For the test vectors not working issue, I made a note in #82.

Primitive/Symmetric/Cipher/Authenticated/AES_GCM.cry
marsella: I think this has something to do with the fact that the wrapper API here isn't actually a functor. I guess the bracket notation only operates directly on a functor, and if I wanted to be able to explicitly specify these types I'd have to make AES_GCM be a functor as well with the additional parameters defined. Maybe this can be done in #80 if we think it's useful.

Primitive/Symmetric/Cipher/Block/AES_specification.cry
```diff
@@ -0,0 +1,64 @@
...
+   type KeySize' : #
+   type constraint (fin KeySize', KeySize' % 64 == 0, KeySize' / 64 >= 2, KeySize' / 64 <= 4)
+
+// The AES implementation uses the `Mode` type to compute other parameters.
+// The mode corresponds directly to the key size:
+// Mode 0 = 128 bits; Mode 1 = 192 bits; Mode 2 = 256 bits
+type Mode = (KeySize / 64) - 2
+
+// Make KeySize accessible outside the module.
+type KeySize = KeySize'
```
weaversa, line 30: I do this all the time too...Is there a more elegant way to get access to parameters?

marsella: I'll ask on our internal channels!

marsella: There is not a more elegant way!. The parameter blocks are anonymous interface modules that are then imported into the named module. But the module system does not automatically export things you've imported, so we have to do it manually.

WeeknightMVP: I'm a poor judge of "elegance", but a modern approach is to specify a (`sub`)`module` for parameterizations whose values are intended to be reused.  Then a user needing the parameter values could `import` the named (`sub`)`module`.

For a parameterized AES module that takes a `KeySize` parameter, an anonymous parameterization like

```cryptol
module Primitive::Symmetric::Cipher::Block::AES128 =
    Primitive::Symmetric::Cipher::Block::AES_specification where
    // This produces a key size of 128
    type KeySize = 128
```

could be refactored to

```cryptol
module Primitive::Symmetric::Cipher::Block::AES128::Parameters where
    // This produces a key size of 128
    type KeySize = 128
```

```cryptol
module Primitive::Symmetric::Cipher::Block::AES128 =
    Primitive::Symmetric::Cipher::Block::AES_specification { Primitive::Symmetric::Cipher::Block::AES128::Parameters }
```

```cryptol
import Primitive::Symmetric::Cipher::Block::AES128::Parameters (KeySize)
import Primitive::Symmetric::Cipher::Block::AES128 (encrypt, decrypt)
```

This approach is a tradeoff -- it is not supported in older Cryptol versions (<3.0), requires another `import` to access an instantiation's parameter values, and at the module level clutters the filesystem, but it reduces redundancy.

A similar concept applies to reusing `parameter` blocks by promoting them to named `interface` (`sub`)(`module`)s.

Issue #80, opened 2024-06-25
Improve code quality on modes of operation


marsella:
We have implementations of most of the modes of operations for block ciphers based on [NIST 800-38A](https://csrc.nist.gov/pubs/sp/800/38/a/final) (and I think some subsequent publications) but they're not very formal. There are a couple things to do here. 

On the presentation side, these aren't literate specs. NIST [is currently revising](https://csrc.nist.gov/news/2023/decision-to-revise-nist-sp-800-38a) the modes of operation document 800-38A, but it's primarily focused on changing guidance and use cases. The updated recommendations for IVs and counter blocks might be relevant for us, but I think they will be more focused on how to choose those fields when encrypting a lot of items, which isn't something Cryptol can capture. It also incorporates some additional versions of CBC mode from an addendum into the spec, but we don't implement those right now. So, if we want to turn this into a literate spec, it's probably reasonable to use the current version of 800-38A as the basis. 

On the engineering side, there are several possible changes to make.
- Currently, most of the modes of operation are defined in terms of a key size; each mode hand-writes the signature for the encrypt and decrypt methods for the block cipher that they'll be used with (others are implemented in terms of a specific block cipher, like AES256). It would be nice to have a more programmatic way to define the mode of operation generically over a block cipher, and then instantiate it for a given block cipher. There is a `Cipher` type that encodes the key size and type of the encryption/decryption methods which is sufficient for almost all modes of operation, except GCM-SIV. We should consider rewriting the modes in terms of Cipher, somehow.
- There are many test vectors in NIST 800-38A that are not implemented. I noticed that CBC/CFB/CTR mode are only instantiated for AES128; there are more AES128 vectors to use for those as well as vectors for other sizes of AES that we could test. There may be others as well, I didn't look very carefully.

amaloz:
@marsella Looks like ECB and OFB are currently missing from the repo. Do we want those?

marsella:
> Looks like ECB and OFB are currently missing from the repo. Do we want those?

We haven't had specific client requests for them, so they've never been added. I don't mind adding them to the issue but they're not high priority to work on right now.

PR #83, opened 2024-06-28
Improve code quality on CTR mode

marsella:
Addresses, but does not complete, #80.

This adds an abstraction for block ciphers and uses it to improve CTR mode. In doing so, I started rearranging the directories in here, to more clearly separate block ciphers from modes of operation from instantiations of a mode with a cipher. Also makes some other clarifications to CTR mode -- improved documentation, tighter type constraints, mild rearranging and renaming of fields to better match the spec.

Some questions:
- I'm not quite sure that this abstraction gives us the right public API for how our clients use modes of operation. I guess they now have a choice of implementing stuff either for a specific instantiation, or by using the CTR functor and filling in the cipher with their own type parameter. This leaves a bit of a middle ground where it's more difficult to specify use of a specific mode with a functor-as-cipher, like "use AES-CTR (without specifying key size)". Thoughts on this?
- @glguy: Do you have an opinion at this point what the best way to document properties and test vectors is to support an easy transition to your docstrings-encoded-batching? If not, I might make a batch file to run them in the interim.
- Scoping: CBC and CFB mode can also be updated to use this paradigm. There are 4 instantiations in the `Modes/` directory that could move to `Instantiations/` -- AES128 and TripleDES with CBC and CFB. I have a slight preference to put these changes in a separate PR so I can spend some time being careful about the modes implementation, but I can do it here, too.


Primitive/Symmetric/Cipher/Block/Tests/TestAES_CTR.cry
```diff
@@ -0,0 +1,81 @@
...
+ *  www.cryptol.net
+ *
+ * References:
+ * [NIST-SP-800-38A]: Morris Dworkin. Recommendation for Block Cipher
+ * Modes of Operation: Methods and Techniques. NIST Special Publication
+ * 800-38A. December 2001.
+ *
+ */
+import Primitive::Symmetric::Cipher::Block::Instantiations::AES128_CTR as AES128_CTR
+import Primitive::Symmetric::Cipher::Block::Instantiations::AES256_CTR as AES256_CTR
```
marsella, line 17: note: we could also do this for AES192, but there doesn't seem to be as much demand for it so I left it out for now. There's no technical reason not to include it though.

mccleeary-galois: Let's either add a note or go ahead and add the test vectors since they are in the Appendix.

mccleeary-galois (approved):
LGTM, I will leave it up to you on whether to comment on 192 or going ahead and adding it now.

Primitive/Symmetric/Cipher/Block/Tests/TestAES_CTR.cry
marsella: Added!

PR #84, opened 2024-07-03
Improve code quality on GCM mode

marsella:
Addresses #78, except for making the spec literate (moved to #87).

This PR aims to improve the code quality, documentation, and NIST spec adherence of GCM mode (and its instantiations for AES). It builds on the paradigms (using `CipherInterface`) added in #83. 

I identified all the issues that I addressed here; if others would like to take a pass of the NIST spec and/or this implementation and identify other areas for improvement, I would welcome that feedback.

---

I have one open / blocking question: There's a duplicate version of AES256-GCM in the repo. It looks like [AES_256_GCM.cry](https://github.com/GaloisInc/cryptol-specs/blob/master/Primitive/Symmetric/Cipher/Authenticated/AES_256_GCM.cry) is used in [the aws-lc-verification repo](https://github.com/search?q=repo%3Aawslabs%2Faws-lc-verification%20AES_256_GCM&type=code). Should keep what I assume is a SAW-optimized version of the spec in this repo? It's not written with respect to the NIST spec (I assume because the NIST spec is not efficient) and I would have to spend some time to audit it and determine whether it's the same as the other version we have. 

I believe there are several other cases where this repo contains duplicate, SAW-optimized versions of algorithms. I'm looking for guidance on what to do in these cases. The options I see are:
1. Remove the SAW-optimized versions from the repo; perhaps move them to another location. Add documentation somewhere at the top level of cryptol-specs to note that this repo aims to add spec-conformant versions of algorithms and they may need to be optimized much more to be used in conjunction with SAW / to actually verify an implementation.
2. Keep the SAW-optimized versions in the repo. Add documentation on each version explaining the use case; add docs at the top level of cryptol-specs to explain the need for multiple versions of things.

Pinging @RyanGlScott and @mccleeary-galois specifically; others also welcome to weigh in.

### DIFF
--- a/Primitive/Symmetric/Cipher/Authenticated/Tests/TestAES_GCM.cry
+++ b/Primitive/Symmetric/Cipher/Authenticated/Tests/TestAES_GCM.cry
@@ -1,39 +1,33 @@
+import Primitive::Symmetric::Cipher::Authenticated::Instantiations::AES128_GCM as AES128_GCM
+import Primitive::Symmetric::Cipher::Authenticated::Instantiations::AES256_GCM as AES256_GCM
 
-// Cryptol instatiation of AES128-GCM and AES256-GCM, and test vectors for each.
-// Copyright (c) 2010-2024, Galois Inc.
-// www.cryptol.net
-// Author: Ajay Kumar Eeralla
+/*
+ Sources for test vectors:
 
-// Test vectors from http://luca-giuzzi.unibs.it/corsi/Support/papers-cryptography/gcm-spec.pdf
+ [GCM Submission]: David A. McGrew and John Viega. The Galois/Counter Mode of
+ Operation (GCM). Submission to NIST Modes of Operation Process. 2004.
+ Accessed from http://luca-giuzzi.unibs.it/corsi/Support/papers-cryptography/gcm-spec.pdf
 
-module Primitive::Symmetric::Cipher::Authenticated::AES_GCM where
-import `Primitive::Symmetric::Cipher::Authenticated::GCM
-import Primitive::Symmetric::Cipher::Block::AES128 as AES128
-import Primitive::Symmetric::Cipher::Block::AES256 as AES256
+ [OPENSSL]: The OpenSSL Project Authors. aesgcmtest.c. 2022.
+ Accessed from https://github.com/openssl/openssl/blob/master/test/aesgcmtest.c
+ */
 
-AES128_GCM_encrypt = GCM_AE `{K=128} {E=AES128::encrypt}
-AES128_GCM_decrypt = GCM_AD `{K=128} {E=AES128::encrypt}
-
-AES256_GCM_encrypt = GCM_AE `{K=256} {E=AES256::encrypt}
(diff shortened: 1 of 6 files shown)


RyanGlScott:
Right, there are various technical reasons why `aws-lc-verification` needs a very bespoke Cryptol specification for AES-GCM. In fact, not all of `aws-lc-verification`'s Cryptol code is contained in `cryptol-specs`, as there are even more Cryptol files in the `aws-lc-verification` repo that make use of things like SMT arrays. One could make the argument that _all_ of these bespoke Cryptol files should live in `aws-lc-verification`, but that's not currently the case.

I could be convinced either way on whether we should keep the bespoke code or remove it. I like the idea of making `cryptol-specs` contain spec-conformant code, so if we _do_ keep the bespoke code, we should prominently advertise why it does something different.

kiniry:
> Right, there are various technical reasons why `aws-lc-verification` needs a very bespoke Cryptol specification for AES-GCM. In fact, not all of `aws-lc-verification`'s Cryptol code is contained in `cryptol-specs`, as there are even more Cryptol files in the `aws-lc-verification` repo that make use of things like SMT arrays. One could make the argument that _all_ of these bespoke Cryptol files should live in `aws-lc-verification`, but that's not currently the case.

Indeed, all bespoke Cryptol and SAW should live somewhere else.  If we are "convinced" that some of it remain, e.g., if we are given permission to share similar work on HIPERSPACE that demonstrates how to assured that such refinement chains are valid, then we can choose to put those either in `cryptol-specs` or some as-of-yet-not-in-existence `cryptol-use` project/repository.

> I could be convinced either way on whether we should keep the bespoke code or remove it. I like the idea of making `cryptol-specs` contain spec-conformant code, so if we _do_ keep the bespoke code, we should prominently advertise why it does something different.

Exactly.  I'd prefer that we do not keep anything that is relevant to a specific client or project in `cryptol-specs`.  Let's get it all moved elsewhere, and notify current and past clients that this is happening.

Primitive/Symmetric/Cipher/Authenticated/Tests/TestAES_GCM.cry
marsella, line 1: note: these are the same test vectors that were previously in `AES_GCM.cry`. The primary changes in this file are that I changed the way we import AES128-GCM and AES256-GCM (previously, they were defined in the same file in which they were tested), and that I added citations.

mccleeary-galois (commented):
LGTM, but I agree with Ryan and Joe that we should make a plan for these SAW-optimized specs for a specific use case to be moved into another repo. Lets plan on chatting more about what to exactly do with that on Monday.

mccleeary-galois (approved):

mccleeary-galois:
@marsella is going to do a quick audit to see if the best course of action for the SAW optimized specs. We will either move them into their respective repos or create a new repo `cryptol-use` or equivalent if it appears to be easier to just dump them into a new repo.

marsella:
I made #88 to follow up on the optimized specs.

PR #90, opened 2024-07-08
ci: add copyright check

### LINKED ISSUE #89: Add copyright checks to CI
Right now we manually add an ad-hoc copyright notice to files. We should add some machine-checked validation to make this more consistent across the repo. The preferred format is a comment at the top of the file that includes the following text:

```
@copyright Galois Inc
@author Name <email>
```

It is reasonable to have additional `@author` tags, and optionally `@editor` tags as more people work on existing files.

At this point, I don't want to modify every file in the repo (and it's not strictly necessary, since copyright is technically optional), so I think we should just start checking this as we go.

- [ ] Add a CI script that checks any modified files for a copyright and author flag

marsella:
Closes #89.

Checks that any files modified, added, etc. in a PR have the appropriate copyright notice on top. You can see an example of a failed and a successful run on the two commits.

### DIFF
--- a/.github/workflows/copyright_check.yml
+++ b/.github/workflows/copyright_check.yml
@@ -1,0 +1,21 @@
+name: Copyright check
+on:
+  pull_request:
+    branches:
+      - master
+
+jobs:
+  copyright-check:
+    runs-on: ubuntu-latest
+    steps:
+      - id: checkout
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - name: "Run copyright script"
+        run: |
+          # The filter excludes files that were removed
+          changed_files=$(git diff --name-only --diff-filter ACMRT ${{ github.event.pull_request.base.sha }} ${{ github.sha }})
+          # NB: this will fail if any files have spaces in the names
+          bash scripts/check_copyright.sh $changed_files
+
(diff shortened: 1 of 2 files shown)


marsella:
It looks like this didn't run the other `ci` check. Any thoughts on why not?

RyanGlScott:
The other CI job [only runs periodically](https://github.com/GaloisInc/cryptol-specs/blob/541077c64472207fb1db0cb20ce258784d566603/.github/workflows/ci.yml#L10), as it uses a `cron` job. As such, it doesn't trigger on every PR. (I'm unclear whether that's desirable or not, but that's my understanding of the current setup.)

marsella:
> The other CI job [only runs periodically](https://github.com/GaloisInc/cryptol-specs/blob/541077c64472207fb1db0cb20ce258784d566603/.github/workflows/ci.yml#L10), as it uses a `cron` job. As such, it doesn't trigger on every PR. (I'm unclear whether that's desirable or not, but that's my understanding of the current setup.)

Oh, I read that as, it runs on push to master, on PR to master, and on the schedule. But I see I've never triggered CI on any of my other PRs, so I guess I'm wrong. That's pretty confusing syntax...

mccleeary-galois (approved):
LGTM, fwiw in the CI overhall of this repo, one of the changes that I had planned was to change it from being a cron job to run on PRs here among other things. You are welcome to change add that as well (keep the cron job as well since that is how we are checking the nightly cryptol build against it right now).

Issue #91, opened 2024-07-15
Convert AES to a literate spec


marsella:
The AES implementation is not currently a literate spec (although it follows closely with [the original NIST spec](https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.197-upd1.pdf). Our gold standard expects all of our popular specs to be Literate.

- [ ] Try to get the original latex document for the spec
- [ ] Convert existing spec to be literate
- [ ] Make sure the instantiations and tests still work (for AES in particular)

marsella:
There is reportedly a literate version of this in the `libvcrypt` (see #31).

PR #92, opened 2024-07-15
Bring AES closer in line with gold standard

### LINKED ISSUE #77: Refactor AES spec to meet "gold standard" requirements
The current AES spec doesn't quite meed the gold standard that we'd like to see for important cryptol specs. This concerns just "plain" AES -- the block cipher itself, not any modes of operation layered on top of it -- as described in [FIPS-197](https://www.nist.gov/publications/advanced-encryption-standard-aes).

Here are the tasks I'm pretty sure need to happen:

- [ ] Convert to literate Cryptol -- determine whether we have access to a latex file of the spec
- [ ] Update module structure to compile all the allowable key sizes by default (a la Kyber)
	- [ ] Uncomment all the test vectors for all the key sizes -- put them in each module
- [ ] Determine whether any other aspects of the structure of the module needs to change to better match the spec
- [ ] Double-check that every data type and function described in the spec has a cryptol corollary
- [ ] Make sure every property in the spec has an equivalent cryptol property
- [ ] Add a batch file to prove or check every property
- [ ] Re-review the gold standard and determine whether anything else needs to change.

marsella:
Closes #77.

This brings our AES implementation closer in line with our gold standard; primary changes are renaming, rearranging, removing unused code, and improving documentation. Note that I punted on making it literate; that task is now captured in #91.


Primitive/Symmetric/Cipher/Block/AES/TBox.cry
marsella, line 1: note: I'm not exactly sure what this was; it looked like maybe an alternate / equivalent version of the S-box? It wasn't used anywhere so I removed it.

Primitive/Symmetric/Cipher/Block/AES/SubBytePlain.cry
marsella, line 1: note: this was consolidated into `Sbox.cry`.

Primitive/Symmetric/Cipher/Block/AES/SubByteSBox.cry
marsella, line 1: note: this was consolidated into `Sbox.cry`.

Primitive/Symmetric/Cipher/Block/AES/Algorithm.cry
```diff
@@ -1,54 +1,78 @@
...
-
-/** AES decryption with an expanded key.
-This is useful if many things will be decrypted with the same key. */
-decrypt : KeySchedule -> [128] -> [128]
-decrypt (kInit, ks, kFinal) ct =
-                          stateToMsg (AESFinalInvRound kFinal (rounds ! 0))
+    // Lines 2-3
+    state0 = kInit ^ msgToState pt
+    // Lines 4-9
+    rounds = [state0] # [ Round::AESRound rk s | rk <- ks | s <- rounds ]
```
marsella, line 57: thought: this is correct but it doesn't look a lot like the Algorithm 1 in the paper. At risk of falling down a nitpicking rabbit hole, I was thinking about combining this file with `Round.cry` and eliminating the `AESRound` and `AESFinalRound` functions, so that it's easier to spot between the spec and the implementation. Thoughts?

Primitive/Symmetric/Cipher/Block/AES/Round.cry
```diff
@@ -1,45 +1,94 @@
...
+InvMixColumns state = GF28::matrixMult m state
     where m = [[0x0e, 0x0b, 0x0d, 0x09] >>> i | i <- [0 .. 3] ]
 
 /** The final inverted AES round */
 AESFinalInvRound : RoundKey -> State -> State
 AESFinalInvRound rk s = rk ^ InvSubBytes (InvShiftRows s)
+
+// This proves.
+subBytesInverts : State -> Bool
+property subBytesInverts s = InvSubBytes (SubBytes s) == s
```
marsella, line 86: note: I tried to annotate every property with how it needs to be called (prove vs check vs check exhaustive), but I didn't make a batch file. Happy to make one if it seems useful.

Primitive/Symmetric/Cipher/Block/AES/Algorithm.cry
marsella, line 1: note: This implementation skips a section (5.3.5) that specifies an alternate, more efficient version of `InvCipher` and `KeyExpansion`. I decided not to add it for the moment for two reasons: I don't think anyone's using `InvCipher` right now anyway, and even if they are, we don't emphasize performance in the gold standard. But if we want it for sake of completeness, I can add.

Primitive/Symmetric/Cipher/Block/AES/Round.cry
mccleeary-galois: You are welcome to start using code fences for this as that feature will be implemented soon in Cryptol.

```
-- @
-- /**
--  * A docstring example
--  *
--  * ```cryptol
--  * :check example
--  * ```
--  */
-- @
```

mccleeary-galois (approved):
LGTM and is a good step to get in. If you want we could start adding some of the comments you have around :check vs :prove into docstrings but happy to have that be a separate thing or decide to not go that route.

Primitive/Symmetric/Cipher/Block/AES/Algorithm.cry
marsella: I ended up doing this in d7a26926.

Primitive/Symmetric/Cipher/Block/AES/Round.cry
marsella: Oh nice. Added!

PR #93, opened 2024-07-15
CI: update version for actions/checkout

marsella:
Following up on [this comment](https://github.com/GaloisInc/cryptol-specs/pull/90#issuecomment-2215069436): I believe the check-build ci run wasn't going because our version of the checkout action was deprecated.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,9 +21,12 @@ jobs:
       run: |
         apt update
         apt install -y git
+
     - name: Checkout
-      uses: actions/checkout@v2
+      uses: actions/checkout@v4
+
     - name: Cryptol Version
       run: cryptol --version
+
     - name: Check Cryptol Files
       run: bash scripts/load_all_cry_files.sh
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,13 +1,9 @@
-name: Cryptol
+name: Cryptol typecheck
 on:
   push:
     branches:
       - master
   pull_request:
-    branches:
-      - master
-  schedule:
-    - cron: "0 10 * * *" # 10am UTC -> 2/3am PST
   workflow_dispatch:
 
 jobs:


RyanGlScott:
I doubt that's the reason (and experimentally, the CI results appear to agree). GitHub Actions will complain loudly about deprecations, but usually it will still run the thing that's deprecated. If something isn't being run at all, that's usually a deeper problem.

marsella:
Hm, this didn't work like I expected. The notation we use suggests that this action should [be triggered by all the specified events], including PRs to master. It actually looks like this action [hasn't run at all in the last 3 months](https://github.com/GaloisInc/cryptol-specs/actions), so there must be something else going on. I'll keep investigating.

RyanGlScott (approved):
That being said, using up-to-date actions is usually a good thing, so I'm happy to land this as a small improvement.

marsella:
Oh, I think I figured it out. Github Actions will apparently [disable scheduled CI jobs](https://docs.github.com/en/actions/using-workflows/disabling-and-enabling-a-workflow) when there has been no activity on a repo for 60 days. In this case, we had it to run both on a schedule and on PR, but once it was disabled for the schedule, it stopped going for PRs, too.

I would propose removing the scheduled run so that it doesn't disappear again; I don't actually see much value in running on a schedule anyway, as long as we have it going for all PRs and pushes to master.

RyanGlScott:
I don't recall why we have the current check set up to run on a schedule, but if it doesn't take too long to run all of the tests, then I would be fine with just running it on every commit. If we identify a need for long-running tests, we can re-introduce them on a schedule later.

marsella:
Right, this specific job only takes 90 seconds historically. If / when we expand CI to actually check properties and do other slower things we can be more discerning.

marsella:
Oh, that makes sense. Hm. I can see an argument for pinning this to something more stable, but that would introduce more maintenance to update with releases of cryptol. Anyway, that seems like an issue to throw back to the maintenance team. I'll make a follow-up issue in this repo.

mccleeary-galois:
> Ah, I think the reason why it's on a schedule is because [it tests against a nightly version of Cryptol](https://github.com/GaloisInc/cryptol-specs/blob/6d7bdc0e71d0aa67d89a75a50f157ff5adcea968/.github/workflows/ci.yml#L17), so I believe the intent is that if someone introduced a Cryptol change which regressed `cryptol-specs`, then the scheduled CI job would be the thing that catches it. That being said, clearly it's not living up to its stated goal, given that we didn't notice this issue for 3 months...
> 
> All of this is to say: feel free to move forward with this PR, although we may want to open a separate issue about reintroducing the scheduled job.

Correct, this is currently how nightly Cryptol is tested against cryptol-specs. We will want to improve this process when we get to the larger CI redo though @marsella.

Issue #94, opened 2024-07-17
Simplify ECDSA module structure


marsella:
Right now, there are two ECDSA implementations in the repo. One uses the curves implemented in `Common/`, and one defines all the legal curves in the same module as the signature implementation. 

In general, we need to figure out what the differences are between these two implementations. But for now, it would help to more cleanly separate the implementations of the curves from the implementations of the signature scheme. 

To do so, I propose adding an interface that lists out the curve operations required to execute ECDSA. This won't be a fully generic `EllipticCurve` interface (since that's a very hard problem) ~and for now it won't even be generic over prime-order Weierstrass curves (the NIST-approved curves for ECDSA, see  [SP 800-186](https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-186.pdf) Table 2). It'll just be a list of the operations we need to execute ECDSA. Then we can parameterize the ECDSA implementation using the new interface and get a bit of separation between the curves and the signing.~ After more careful review of the requirements for ECDSA and ECDH, I decided that we should just have a generic version of prime-order short-weierstrass curves, since almost all the non-deprecated curves for all the specs we want to implement fall under that category[^bp].

[^bp]: There's an important difference between the NIST-standardized curves and the brainpool curves, which are also approved. All the NIST algorithms are optimized for curves with `a=-3` but the brainpool curves don't have this restriction, I think. If we want to expand support to the brainpool curves later we'll have to refactor to actually have an EC interface that can be instantiated by either bp or nist curves, then parameterized by the appropriate parameter sets. But I think that's a problem we can address later.

~My expectation is that once the curves and the signatures are separated, and perhaps once we start implementing ECDH, we can figure out whether there's a better, more general interface for curves that would be useful in a wider variety of algorithms.~
Now my expectation is that we will do this in several PRs. First:

- [x] Define a reasonable prime-order elliptic curve interface - see #96.
- [x] Port the tests on elliptic curves from `ecdsa.cry` to the interface.

Then:
- [x] Update one of the ECDSA implementations to use the new interface
- [x] Add instantiations of ECDSA for the interesting curves
- [x] Make a tests directory for ECDSA that includes test vectors from the specs

Finally:
- [x] Clean up all the remaining elliptic curve and ECDSA implementations (see #119)

marsella:
Okay, after some continued exploration, I determined a few things:
- There are 2 different ECDSA implementations
[...]

PR #96, opened 2024-07-19
Add spec-adherent elliptic curve implementation

marsella:
Addresses part of, but does not complete, #94.

This pulls a bunch of optimized elliptic curve routines from [another implementation](https://github.com/GaloisInc/cryptol-specs/blob/master/Primitive/Asymmetric/Signature/ECDSA/ECDSA.cry) and adds a lot of documentation and property checking throughout.

I ended up pulling algorithms from multiple sources, not just the specification that defines the elliptic curve parameters, because that file only loosely defines the basic EC group operation, and not in any kind of optimized way. The old scalar multiplication algorithm in this file was from that spec, and it took 5+ minutes to _check_ a single instance of scalar multiplication. Now we can check 100 in <1min. 

Some notes:
1. It's not tested at all! I mean, it is in the sense that Sean tested the algorithms in the other ECDSA implementation, but this particular module has nothing beyond the few properties in here. I searched around and didn't find any EC operation test vectors; NIST only seems to provide them for the algorithms that use ECs (ECDSA, ECDH, etc.). See #97.
2. I'm not sure if this is the right API since I haven't tried to use it yet. Specifically I hide all the projective representation stuff behind a `private` block and only reveal operations on affine points. We might want this to be different but I figured we can faff around with it once we have a concrete implementation.
3. This is intended to supersede the three other implementations of elliptic curves (`EC_P384.cry`, not used for anything right now; `p384_{ec_mul, ec_point, field}`, which is used in the `ecdsa.cry` implementation; and the curve implementation in `Signature/ECDSA/ECDSA.cry`).

I decided to stop here because it's already a fairly large PR and I didn't want to add a whole ECDSA refactor + test vectors on top of this. I updated #94 to better capture my plan now. Upcoming PRs include
- Add the ec-only test vectors from one of the ecdsa files
- Use this in an ECDSA implementation
- Add twin-mult and test vectors.


Common/EC/PrimeField/PFEC.cry
```diff
@@ -1,57 +1,381 @@
...
+        {x = r32, y = r37, z = r27} /* 38: Rx <- t1; Ry <- t2, Rz <- t3 */
+      where (t1, t2, t3, t4, t5, t6) = if T.z == 1
+              then (S.x, S.y, S.z, T.x, T.y, T.z) /* 1: we set t6 = 1 to make L25 work */
+              else (r5, r7, S.z, T.x, T.y, r3)
+
+            r3 = T.z                /* 3: t6 <- T_z */
+            r4 = r3 ^^ 2            /* 4: t7 <- t6^2 */
+            r5 = S.x * r4           /* 5: t1 <- t1 * t7 */
+            r6 = r3 * r4            /* 6: t7 <- t6 * t7 */
+            r7 = S.y * r6           /* 7: t2 <- t2 * t7 */
```
marsella, line 293: @weaversa I pulled most of these implementations from [your 2018 ECDSA implementation](https://github.com/GaloisInc/cryptol-specs/commit/b07136437b38789a9d5ad85501c00780bb7aa24a), but I noticed the `ec_add` function was missing all the conditionals for when `T_z != 1`. Do you remember why you didn't include those? 

My best guess is that, if we assume we only are using this as a sub-call of `ec_mult`, then we know that we normalized the right-hand-side element (in L10-12 of `ec_mult`) so the z-coord will always be 1.

(Routines are from [[MATH-2008](https://citeseerx.ist.psu.edu/document?repid=rep1&type=pdf&doi=e1b05e600bfcdeecf8555bc9948483c5fbbdd478)]).

Common/EC/PrimeField/PFEC.cry
```diff
@@ -1,57 +1,376 @@
...
+    Infinity -> point2
+    Affine x1 y1 -> case point2 of
+        Infinity -> point1
+        Affine x2 y2 -> if (x1 == x2) && (y1 == -y2) then Infinity
+            else Affine x y
+            where  // all arithmetic operations are mod p
+                x = lambda ^^ 2 - x1 - x2
+                y = lambda * (x1 - x) - y1
+                lambda = if x1 == x2 then (3 * x1 ^^ 2 + a) %/ (2 * y1)
                     else (y2 - y1) %/ (x2 - x1)
```
marsella, line 180: note: this is the inefficient affine version from the spec. Any reasonable person will probably want to use the fast projective version. Maybe the appropriate thing is to keep this implementation somewhere private and prove it's equivalent to the alternate `ec_add`.

Common/EC/PrimeField/PFEC.cry
```diff
@@ -1,57 +1,376 @@
...
+ * notes that it can be computed efficiently using the addition rule and a
+ * variant of the double-and-add algorithm.
+ * In this implementation, we use a variant from [MATH-2008] (see `ec_mult`).
+ */
 scmul : Integer -> Point -> Point
-scmul m xy =
-    if m == 0 then oo
-     | m == 1 then xy
-     | even m then scmul (m / 2) (ec_add xy xy)
-              else ec_add xy (scmul (m - 1) xy)
```
marsella, line 48: note: this used to be an incredibly inefficient affine version. I couldn't actually find this specified anywhere beyond a line in SEC1 that said, effectively "you can use double-and-add for this" (without actually describing the algorithm). I could keep it and prove it equivalent but I'm not sure that adds value when the slow version isn't even specified.

Common/EC/PrimeField/PFEC.cry
weaversa: I just plain don't remember, though I appreciate your suggestion that I was doing something clever. What you have here (following the specification) is better.

Common/EC/PrimeField/PFEC.cry
```diff
@@ -1,57 +1,376 @@
...
-    type P : #     //  The prime modulus
-    type constraint (fin P, P >= 5)
-    n  : Integer   //  Prime order of generator G
-    b  : Z P       //  Curve coefficient
-    Gx : Z P       //  Curve coordinates
-    Gy : Z P       //    of generator point
+    /**
+     * Field size `P`.
+     * P must be an odd prime power and is not divisible by 3.
+     * I don't think there's a way to enforce "prime power" in the type
```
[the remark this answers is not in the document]
mccleeary-galois: This does allow you to do things like:

```
id : {n, a} (prime n, fin a) => [n^^a] -> [n^^a]
id a = a
```

Issue #98, opened 2024-07-26
Add a Hash interface


marsella:
Many algorithms, like ECDSA, are generic over the hash function they use as long as it satisfies certain properties (e.g. appropriate-length hash output). However, right now there is no unifying interface for hash functions. We should write an interface so that other algorithms can be written generically.

Some useful things we'll want in it, from implementing ECDSA:
- An accessible `hashlen` (the length of the output)
- the `Hash` function itself
- I think most hash functions have a constraint on the maximum message size. There should probably be a type constraint as part of the interface that can be used to restrict inputs on functions that call the hash.

There may be other useful things to put in the interface, too.

- [x] Implement an interface
- [x] Instantiate it with the SHA2 functions as an example
- [x] Consider instantiating with other useful hashes as well, or make follow-up issues to do so.
- [x] Update ECDSA to use the interface (as an example. If this is too hard, make a follow-up issue.)
- [x] Identify other algorithms that could use it and make follow-up issues.

marsella:
Notes on the type constraint idea:
- Interfaces cannot contain generic / fill-in-the-blank constraints. They can contain constraints that must be true for all implementations of the interface
- A brief pass suggests that many hash functions use the specific constraint that a message must be no more than `n` bits, for some `n`. We might be clear to encode this upper bound as a type on the interface and then define the constraint at the interface level (instead of the implementation level).

weaversa:
> Notes on the type constraint idea:
> 
> * Interfaces cannot contain generic / fill-in-the-blank constraints. They can contain constraints that must be true for all implementations of the interface
> * A brief pass suggests that many hash functions use the specific constraint that a message must be no more than `n` bits, for some `n`. We might be clear to encode this upper bound as a type on the interface and then define the constraint at the interface level (instead of the implementation level).

In a recent take we have the following in a parameter block:
```
  type UpperLimit : #
  hash : {a} (fin a, UpperLimit >= width a) => [a] -> [DigestSize]
```

This is workable, but I'd love to see an improvement. In the back of my head I've always thought it would be nice if Cryptol could accept type constraints at the parameter level, along with the current values and types (natural numbers).

PR #99, opened 2024-07-29
Add ECDSA signing + verification

marsella:
This addresses the second part of #94.

- It adds a spec for ECDSA signing + verification that can be parameterized by a suitable curve.
- It adds an interface that defines the basic set of elliptic curve operations we need for this application (and hopefully other applications, too)
- It instantiates ECDSA for P384 and SHA-256 and evaluates the test vector from RFC 6979.

Remaining work (possibly to be punted to another PR):
- [x] Fix the regression in all the files that I accidentally broke
- [x] Decide whether to add more test vectors -- see #104 
- [x] Review spec more carefully and make sure this meets our gold standard of literacy
- [x] Review docs again and make sure appropriate citations are everywhere they belong.

Future work that will definitely be in separate PRs:
- Implement twin-mult to make verification faster -- see #103 
- Remove the other versions of ECDSA -- see remaining checkboxes in #94


Common/utils.cry
```diff
@@ -67,24 +67,40 @@ half x = fromInteger (if isEven xint then xint/2 else ((xint + `p) / 2))
 half_correct : {p} (fin p, p >= 3, isOdd p) => Z p -> Bit
 property half_correct x = half x + half x == x
 
-mp_mod_inv : {n} (fin n, n >= 1) => Z n -> Z n
+/**
```
weaversa, line 70: Is this needed now that we have `/.` for `Z` types?

Common/EC/ECInterface.cry
marsella, line 1: note: this doesn't really have an equivalent in any spec. It's more of a mathematical concept than a specified algorithm. Some instantiations of this will be well-specified, and are cited appropriately, but the docs in here are very light.

Thoughts on whether there's more documentation to put in here would be helpful.

Primitive/Asymmetric/Signature/ECDSA/specification.cry
```diff
@@ -0,0 +1,166 @@
...
+
+        // Step 4.
+        k_inv = mp_mod_inv k
+
+        // Step 5.
+        R = EC::scmul (fromZ k) EC::G
+
+        // Steps 6-8. This fails (returns `None`) if `R` is the point at
+        // infinity.
+        maybe_r = EC::xCoordModOrder R
```
marsella, line 78: question: would it be safe to unwrap this? By definition `R` is a scalar product of the base point `G` (where `G` generates a large prime-order subgroup of the curve). Is there a `k` such that `[k]G` is the point at infinity? I have to go study more elliptic curve properties to know.

Common/utils.cry
```diff
@@ -67,24 +67,40 @@ half x = fromInteger (if isEven xint then xint/2 else ((xint + `p) / 2))
...
+    (u'', _, x1'', x2'') = while outtercond outterbody (fromZ c, `p, 1, 0)
 
-// Note, this property will only hold when the modulus is prime
-mp_mod_inv_correct : {a} (fin a, a >=2) => Z a -> Bit
+/**
+ * The inverse must be correct.
+ * This property will only hold when the modulus is prime!
+ * It must be `:check`ed.
+ */
+mp_mod_inv_correct : {a} (fin a, prime a, a >=2) => Z a -> Bit
```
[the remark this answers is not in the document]
marsella: Not needed since I switched to using the built-in function.

Primitive/Asymmetric/Signature/ECDSA/specification.cry
marsella: Right! Thank you. After thinking about it, I decided I don't mind the case statement below, but I'll add documentation to explain why the `None` case should be impossible here.

weaversa: This may be a good use of Cryptol's safety checking capabilities (`:safe`).

marsella:
Flagging #102 as a related change that I'd like some input on!

Common/EC/PrimeField/PFEC.cry
```diff
@@ -295,6 +295,16 @@ property scmulIsClosed m = valid_point (m ~* G)
  */
 property scmul_commutes m m' = affineEq (m ~* (m' ~* G)) (m' ~* (m ~* G))
 
+/**
+ * Compute twin multiplication.
+ *
+ * There's a more efficient routine to implement this functionality, but we
+ * haven't implemented it here yet. [MATH-2008] Routine 2.2.12.
```
mccleeary-galois, line 302: Do we have a ticket tracking improvements like this?

Primitive/Asymmetric/Signature/ECDSA/specification.cry
```diff
@@ -0,0 +1,231 @@
...
+
+        // This fails (returns `None`) if `r` or `s` are invalid.
+        maybe_rs = case maybe_r of
+            // Step 11.
+            Some r -> if (r != 0) && (s != 0) then Some (r, s) else None
+                // Step 9. The `mod n` is implicit here because these are all
+                // of type `Z n`.
+                where s = k_inv * (e + r * d)
+
+            None -> None // Impossible!
```
mccleeary-galois, line 144: Should we use `error` here since it is truly impossible? or is `None` safe?

mccleeary-galois (approved):
Not an expert, but generally LGTM and follows the sources cited.

Primitive/Asymmetric/Signature/ECDSA/Tests/ECDSA_P384.cry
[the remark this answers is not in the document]
marsella: Discussed offline: yes, I'll do this in a separate PR. See #104.

Common/EC/PrimeField/PFEC.cry
marsella: Oops, no. I made #103 for twin-mult.

Primitive/Asymmetric/Signature/ECDSA/specification.cry
marsella: Well, the function returns an option either way so it's not a problem to keep the none case around. In general I prefer handle-able errors (options, results) over unhandle-able ones (panics, errors) but I don't know if that aligns with the general programming philosophy in Cryptol.

mccleeary-galois: I prefer not using it as well, was just wondering if the `Impossible` comment meant that hitting this path was one that warranted it.

mccleeary-galois (approved):

Primitive/Asymmetric/Signature/ECDSA/Tests/ECDSA_P384.cry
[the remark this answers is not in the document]
kiniry: Thank you for being careful about filing new issues to chase up this kind of thing.

Common/EC/ECInterface.cry
kiniry: I would point in the header via a `@see` semantic property at least one canonical source reference that we/clients use to teach and learn about ECs.  This can point to a book, a paper, a tutorial, etc.

kiniry: A `README.md` is warranted in `Common/EC`.

Primitive/Asymmetric/Signature/ECDSA/specification.cry
kiniry: Why is this called `specification` rather than `Specification`?

Primitive/Asymmetric/Signature/ECDSA/specification.cry
kiniry: I believe that @weaversa has a strong opinion about such?

kiniry (approved):
That's all of my comments for the moment.  The only thing I do not understand at a big-picture level is how/when these detailed non-literate specs get integrated into the full literate spec.

Primitive/Asymmetric/Signature/ECDSA/specification.cry
marsella: @weaversa I hadn't seen that previously, thank you for the suggestion!

@kiniry I was matching a convention I saw elsewhere in the repo, but I'm not tied to it.

Common/EC/ECInterface.cry
```diff
@@ -0,0 +1,69 @@
...
+    twin_mul : Integer -> Point -> Integer -> Point -> Point
+
+    /**
+     * ECDSA requires a routine to extract the x-coordinate of the affine
+     * representation of a `Point` and convert it from a field element an
+     * integer mod the order of the curve (`n`). This is not defined and
+     * returns `None` for the point at infinity.
+     *
+     * For NIST-standardized curves, this should match the conversion
+     * routine in [SP-800-186] Appendix F.1.
```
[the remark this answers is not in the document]
marsella: Oh, I haven't seen this before. In most of the files I've been putting a list of references at the top that has the full citation + URL. I see I forgot that here. Maybe I'll update my references to include the `@see` annotation so it's machine-readable without losing the whole citation info in case of url rot, etc. (and I'll add one here).

Common/utils.cry
```diff
@@ -67,24 +67,40 @@ half x = fromInteger (if isEven xint then xint/2 else ((xint + `p) / 2))
...
         (v', x2') = while innercond innerbody (v, x2)
 
-    (u'', _, x1'', x2'') = while outtercond outterbody (fromZ c, `n, 1, 0)
+    (u'', _, x1'', x2'') = while outtercond outterbody (fromZ c, `p, 1, 0)
 
-// Note, this property will only hold when the modulus is prime
-mp_mod_inv_correct : {a} (fin a, a >=2) => Z a -> Bit
+/**
+ * The inverse must be correct.
+ * This property will only hold when the modulus is prime!
```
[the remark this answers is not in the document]
marsella: I made #100 to fix this -- inverses in the integer mod groups are now a built-in function in cryptol, so I'm planning to take this out entirely.

Primitive/Asymmetric/Signature/ECDSA/unconstrained_spec.cry
```diff
@@ -0,0 +1,222 @@
...
+ *
+ * Outputs:
+ * Accept (True) or reject (False) the signature over `M` as originating from
+ * the owner of public key `Q`.
+ *
+ * In all inputs and outputs, `n` is the order of the base point `G` for the
+ * elliptic curve specified in the `PFEC` interface.
+ */
+verify : {m} (fin m, width m < 64) => [m] -> (Z EC::n, Z EC::n) -> EC::Point -> Bool
+verify M (r, s) Q = inputsInRange && rMatches
```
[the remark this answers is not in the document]
marsella: Parameter ordering matches the NIST spec. And yes, due to asymmetric keys, the types of the signing and public keys are different so `sign` and `verify` have different signatures. I can add some type aliases to make it more obvious what each piece is at the type level and illustrate the parallels between the two functions -- `SigningKey` and `PublicKey` and `Signature`.

Primitive/Asymmetric/Signature/ECDSA/unconstrained_spec.cry
```diff
@@ -0,0 +1,222 @@
...
+signAndVerifyIsCorrect : {m} (fin m, width m < 64) => [m] -> Z EC::n -> Z EC::n -> Bool
+property signAndVerifyIsCorrect M d k = verification
+    where
+        Q = publicKey d
+        signature = sign M d k
+        verification = case signature of
+            Some (r, s) -> verify M (r, s) Q
+            None -> False
+
+private
```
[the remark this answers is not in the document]
marsella: Actually, the truncation and integer conversion is specific to ECDSA, it's just the `Hash::sha` line that will move to be part of the generic `Hash` interface. I pulled it into a helper function because we do the same message processing in both sign and verify and it seemed good to avoid code duplication, but I am happy to paste it back inline if that's confusing.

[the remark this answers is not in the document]
marsella: See #98 for the general hash interface issue.

Primitive/Asymmetric/Signature/ECDSA/specification.cry
weaversa: Well...I agree that the `None` case is impossible and so in my own implementation I `unwrap` the result. However, I don't think it's bad to be overly verbose about the error cases. As written here it's clear something bad that can happen if misimplemented, whereas folks may skip over my `unwrap` take.

Common/EC/ECInterface.cry
marsella: Added references, a readme, and a bonus readme for ECDSA because there are simply too many implementations to keep straight.

mccleeary-galois (approved):

PR #102, opened 2024-07-31
ecdsa: add hash length constraint #94

marsella:
Making this PR because I want input on how to proceed.

I did not realize this previously, but [the DSA standard](https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.186-5.pdf) has a note that requires the security level of the elliptic curve to be greater than the security level of the hash function (see Section 6.1.1 on page 30). This is reasonable in theory -- it would be an easy mistake to think you get security up to the parameter size of the elliptic curve but then have your hash function cut it much shorter.

I'm not sure that practical uses in the wild actually adhere to this requirement. For example, Rust Crypto has a method that [associates a preferred digest to a curve](https://docs.rs/p256/latest/p256/struct.NistP256.html#impl-DigestPrimitive-for-NistP256) but also allows [signing a digest directly](https://docs.rs/ecdsa/latest/ecdsa/struct.SigningKey.html#method.sign_digest_recoverable), and the latter I believe allows any length digest.

As another example, [RFC 6979](https://datatracker.ietf.org/doc/html/rfc6979) includes test vectors for many combinations of curves and hash functions, including combinations that are not allowed with this constraint. Indeed, CI for this PR should fail because the first test vector I made (P384 + SHA256) does not meet the requirement.

I would appreciate feedback from internal and external folks about whether we want to encode this requirement into the standard.

--- 
If we do decide to encode this, we might want to increase the priority of #98 so that it's easier to instantiate valid curve+hash pairs.


Primitive/Asymmetric/Signature/ECDSA/specification.cry
```diff
@@ -62,170 +66,16 @@ interface constraint (width EC::n >= 224)
...
-            // We further convert it to an element in `Z n` to support the modular
-            // operations later in the protocol.
-            e = fromInteger (toInteger e')
+// Documentation for the public interface of this API can be found in
+// `unconstrained_spec.cry`.
+sign = USpec::sign
+verify = USpec::verify
+publicKey = USpec::publicKey
 
+// NB: removed property
```
marsella, line 81: question: Should the "sign + verify works" property be in this module or in the unconstrained version?

Primitive/Asymmetric/Signature/ECDSA/specification.cry
```diff
@@ -62,170 +66,16 @@ interface constraint (width EC::n >= 224)
...
-    hashAndTruncate M = e
-        where
-            H = Hash::sha M
-            e' = take`{min (width EC::n) Hash::digestSize} H
-            // Cryptol's default bitstring-to-integer `toInteger` conversion
-            // matches the routine specified in [FIPS-186-5] Appendix B.2.1.
-            // We further convert it to an element in `Z n` to support the modular
-            // operations later in the protocol.
-            e = fromInteger (toInteger e')
+// Documentation for the public interface of this API can be found in
```
marsella, line 75: question: Are we okay with omitting all the docs here? Should top-level docs live here instead of in the unconstrained version? (I left them in unconstrained because it seemed nice to have the docs next to the concrete implementation). I didn't want to duplicate them to avoid drift between the two versions.

Primitive/Asymmetric/Signature/ECDSA/unconstrained_spec.cry
```diff
@@ -0,0 +1,222 @@
+/*
+ECDSA as specified in [FIPS-186-5] Section 6.
+
+⚠ This implementation is complete except for constraints on the size of the
+input parameters ([FIPS-186-5] Section 6.1.1)! Using this implementation will
+allow invalid combinations of parameters. Consider using `specification.cry`
+instead.
```
marsella, line 7: note: These lines are the only new thing in this doc. Everything else was ported directly from `specification.cry`.

PR #106, opened 2024-08-05
Add twin multiplication for prime field curves

### LINKED ISSUE #103: Add efficient twin-mul implementation for NIST P-curves
Currently, twin multiplication on elliptic curves (`[a]P + [b]Q` for scalars `a`, `b` and curve points `P` and `Q`) is implemented naively for the NIST P-curves. We should implement the optimized algorithm in [the NIST EC doc](https://citeseerx.ist.psu.edu/document?repid=rep1&type=pdf&doi=e1b05e600bfcdeecf8555bc9948483c5fbbdd478) (Routine 2.2.12

Note that this issue is for the `PFEC` implementation. There is a sample implementation of twin-mul in `Primitive/Asymmetric/Signature/ECDSA/ECDSA.cry` that we should draw from. There's also an implementation in `Common/EC/p384_ec_mul.cry` but it looks pretty different from the coding style in the rest of the implementation.

- [ ] Implement twin mult!
- [ ] Do a quick benchmark to see how much faster it is?

marsella:
Closes #103

This adds an efficient utility function for prime field elliptic curves that's used in decryption for ECDSA. Twin multiplication is the operation 
$` f(x, S, y, T) = [x]S + [y]T `$

I am open to debate about whether this is worth including in this repo. It has two parts: twin multiplication itself and an optimized method to normalize four points at once. Neither of these are standardized by NIST or any other body, as far as I can tell (they're included in the "optimized routines for NIST curves" doc). Neither of them are necessary for correct functionality -- we can always do the naive thing of two scalar multiplications and an addition. It's also not intuitive at all (to my eyes, at least).

### DIFF
--- a/Common/EC/PrimeField/PFEC.cry
+++ b/Common/EC/PrimeField/PFEC.cry
@@ -1,46 +1,46 @@
 /*
- NIST-standardized Prime Field Elliptic Curves in short-Weierstrass form.
+NIST-standardized Prime Field Elliptic Curves in short-Weierstrass form.
 
- This contains an implementation of elliptic curve operations for curves
- in short-Weierstrass form, defined over the finite field GF(P), where P
- is an odd prime power (short-Weierstrass form can also be used over the
- binary field GF(2^m), but curves of this form were deprecated in
- [SP-800-186] Section 3.3 and are not supported by this implementation).
+This contains an implementation of elliptic curve operations for curves
+in short-Weierstrass form, defined over the finite field GF(P), where P
+is an odd prime power (short-Weierstrass form can also be used over the
+binary field GF(2^m), but curves of this form were deprecated in
+[SP-800-186] Section 3.3 and are not supported by this implementation).
 
- This implementation has several restrictions that tie it specifically to
- the NIST-standardized curves:
- - it fixes the curve parameter `a = -3`;
- - it fixes the cofactor `h = 1`;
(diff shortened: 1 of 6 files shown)


Common/EC/PrimeField/PFEC.cry
```diff
@@ -314,6 +314,53 @@ private
...
+    }
+    /**
+     * Optimized routine to affinify and projectify four points at once
+     * (or do nothing, if any of the points are the point at infinity).
+     *
+     * This optimization is designed for use in `ec_twin_mult` and should not
+     * be used in other settings without checking that its behavior is suitable.
+     * [MATH-2008] Section 2.2, Note 8.
+     */
+    twin_normalize: TwinPoints -> TwinPoints
```
marsella, line 334: note: I tried writing a property to prove this works as expected, but my random-ec-point generator creates affine points, which, when converted to projective representation, are already normalized. It didn't seem very useful to prove that only normalizing normalized points works, but I don't have a good way to generate random, non-normalized projective points.

mccleeary-galois: maybe we just rely on discarding of tests here with a guard on the property, it also doesn't hurt to test that when given normalized points it is a no-op.

mccleeary-galois (approved):
LGTM, followed along with the spec cited.

PR #107, opened 2024-08-05
Remove uses of coerceSize

### LINKED ISSUE #101: Clean up uses of `coerceSize`
Now that https://github.com/GaloisInc/cryptol/pull/1713 has been merged, some use cases of `coerceSize` should be able to be removed from `cryptol-specs`, particularly in ML-KEM, e.g. https://github.com/GaloisInc/cryptol-specs/blob/049f57d681c3a45469d8b533a944be9dd42988d7/Primitive/Asymmetric/Cipher/ML-KEM/specification.cry#L198

marsella:
Closes #101 

This removes all the uses of `coerceSize`. It also remove a bunch of trailing whitespace, because my editor does that automatically, and adds a copyright notice.

~~This is on hold for now because as mentioned in the issue, it depends on a nightly-only feature of cryptol.~~ Now ready for review after the release of Cryptol 3.2.0!


Common/ntt.cry
```diff
@@ -87,21 +87,17 @@ roots = iterate ((*) (r * r)) 1
...
+ntt_r depth a
+  | n == 0 =>  a
+  | n > 0 => butterfly depth even odd
     where
-      (lft, rht) = shuffle (coerceSize a)
-      even = ntt_r`{max 1 n - 1} (depth + 1) lft
-      odd = ntt_r`{max 1 n - 1} (depth + 1) rht
+      (lft, rht) = shuffle a
+      even = ntt_r`{n - 1} (depth + 1) lft
+      odd = ntt_r`{n - 1} (depth + 1) rht
```
marsella, line 100: question: Do I need to keep the max 1 in here? What happens if this gets called with `n=1`? Don't we want to hit the 0 case eventually?

marsella: I learned a Haskellism today! The association is `(max 1 n) - 1`. Now that we have numeric constraint guards to enforce that `n >= 1`, the max's can go. I'll remove throughout.

Common/ntt.cry
```diff
@@ -115,13 +116,14 @@ shuffle a =
 /**
  * Perform the butterfly operation.
  */
-butterfly : {n} (fin n, n > 0) => Integer -> [n]Fld -> [n]Fld -> [2 * n]Fld
```
yav, line 118: Why did we make the type of `butterfly` *less* general here (i.e., it seems looks like the function makes sense for any length > 0, rather than just powers of 2)

yav (approved):

Common/ntt.cry
marsella: Oh gosh, good question. I don't remember what I was thinking here, but probably I was just iterating on trying to understand the implementation and accidentally left this in. I'll change it back, thanks for catching this!

PR #108, opened 2024-08-05
Fix docstrings

### LINKED ISSUE #105: Fix docstrings for AES & friends
I've added lots of properties in the last few months that don't use the conventions of the new `:check-docstrings` command. I want to go back and fix all of those. Goal is that all properties have a docstring with a `check`/`prove`/`exhaust` command on it.

- [ ] Fix properties in AES
- [x] Fix properties in modes of operation (CBC, CTR, CFB, GCM)
- [ ] Remove batch files (make sure all the things called are now contained in docstrings)

marsella:
Closes #105.

This fixes some docstring checks and proves to use the correct notation, instead of an outdated version, and updates some comments about provability to actually have a proving docstring. It also cleans up some parameterized imports in AES that didn't need to be parameterized.


mccleeary-galois (approved):
LGTM, `:check-docstring` seems to be working well @glguy on these. One thing of note is for extremely large `:check-docstring` processes there is no notion if it is just hanging or taking a long time to compute. I know that this is an issue for any property check that takes a significant amount of time but this is a bit extended by the fact that all of the checks need to complete before output is produced. No action needed on this comment at the moment just something to note.

Primitive/Symmetric/Cipher/Authenticated/aes_gcm.bat
glguy, line 1: I checked that all of these properties were distributed to a doctest.

Primitive/Symmetric/Cipher/Authenticated/GCM.cry
```diff
@@ -113,14 +113,28 @@ GCM_AD key iv ct aad tag =
     T' = MSB`{T} (GCTR CIPHk J0 S)
 
 /**
-* Property demonstrating equivalence between `mult` and `•`.
-*/
+ * Property demonstrating equivalence between `mult` and `•`.
+ * This takes more than 25 minutes to `:prove`.
+ * ```repl
+ * :check dotAndMultAreEquivalent
```
glguy, line 119: That it's not obvious if it's most appropriate to prove, exhaust, or check a property like this seems like it means that having to put explicit repl fences on properties isn't a completely redundant exercise.

glguy (approved):
Everything looks right except:

- `GC28.cry` has a couple of code fences using `cryptol` instead of `repl` that I think should be changed. I checked and those properties do pass unless that module is out of scope for things change.

These changes help demonstrate that it can be confusing to know which modules are functors and which are not and therefore which modules should be able to have their docstrings tested.

Primitive/Symmetric/Cipher/Authenticated/GCM.cry
marsella: Agreed. It seems difficult to me for an outsider to come in and tell by looking how they should interact with properties.

PR #114, opened 2024-08-08
Add preliminaries for SHA2

### LINKED ISSUE #109: Implement preliminaries for SHA2
This issue is an attempt to outline a reasonably-scoped first PR for implementing the SHA2 algorithms from [FIPS 180-4](https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.180-4.pdf). 

Implement the preliminary functions and properties from sections 1-4 of FIPS 180-4.

- [x] Properties that ROTL and ROTR are equivalent to the >>> operators in cryptol (maybe make explicit functions for them) (2.2.2)
- [x] Properties showing all the example representation conversions in 3.1
- [x] Equivalence relationship for ROTL/R in 3.2 # 6
- [x] Define a nice abstraction for the logical functions in 4.1.2 and 4.1.3. This will also require deciding what our parameters are.

marsella:
Closes #109.

This moves all the existing SHA work (which is a mostly imperative implementation for a specific use case) to a separate directory and defines all the preliminary notation in sections 1-4 of the [FIPS 180-4 spec](https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.180-4.pdf).

I'm trying to just add parameters as I need them, so this is definitely not complete yet, but it should contain all the properties and definitions from the noted sections.

### DIFF
--- a/Primitive/Keyless/Hash/SHA2/Specification.cry
+++ b/Primitive/Keyless/Hash/SHA2/Specification.cry
@@ -1,0 +1,36 @@
+/*
+SHA2 specification.
+
+References
+[FIPS-180-4]: National Institute of Standards and Technology. Secure Hash
+  Standard (SHS). (Department of Commerce, Washington, D.C.), Federal
+  Information Processing Standards Publication (FIPS) NIST FIPS 180-4.
+  August 2015.
+  @see https://doi.org/10.6028/NIST.FIPS.180-4
+
+*/
+module Primitive::Keyless::Hash::SHA2::Specification where
+
+
+/**
+ * Circular rotate left operation.
+ * [FIPS-180-4] Section 2.2.2.
+ */
+ROTL : {w} (fin w, w > 0) => Z w -> [w] -> [w]
+ROTL n x = x <<< (fromZ n)
+
+/**
+ * Circular rotate right operation.
+ * [FIPS-180-4] Section 2.2.2.
+ */
+ROTR : {w} (fin w, w > 0) => Z w -> [w] -> [w]
+ROTR n x = x >>> (fromZ n)
+
+/**
+ * Right shift operation.
+ * [FIPS-180-4] Section 2.2.2.
+ */
+SHR : {w} (fin w, w > 0) => Z w -> [w] -> [w]
+SHR n x = x >> (fromZ n)
+
+
--- a/Primitive/Keyless/Hash/SHA2/Specification.cry
+++ b/Primitive/Keyless/Hash/SHA2/Specification.cry
@@ -136,6 +136,44 @@ sigma1 x
     | w == 32 => ROTR 17 x ^ ROTR 19 x ^ ROTR 10 x
     | w == 64 => ROTR 19 x ^ ROTR 61 x ^ ROTR  6 x
 
-
-
-
+/**
+ * The SHA family uses a sequence of constant `w`-bit words, which represent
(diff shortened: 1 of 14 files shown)


Primitive/Keyless/Hash/SHA2/Specification.cry
```diff
@@ -0,0 +1,187 @@
...
+module Primitive::Keyless::Hash::SHA2::Specification where
+
+parameter
+    /**
+     * Word size.
+     * The specification defines words to be either 32 or 64 bits.
+     * [FIPS-180-4] Section 1, Figure 1.
+     */
+    type w : #
+    type constraint (fin w, w % 32 == 0, 1 <= w / 32, 2 >= w / 32)
```
marsella, line 29: note: in the previous PR we discussed not encoding mathematically unnecessary constraints that are only part of the spec for consistency / security reasons. I'm hard-coding this here because the spec doesn't give instructions for arbitrary word sizes, especially in the constants & so on.

Primitive/Keyless/Hash/SHA2/Specification.cry
```diff
@@ -0,0 +1,187 @@
...
+
+/**
+ * Circular rotate left operation.
+ * [FIPS-180-4] Section 2.2.2 and Section 3.2 #5.
+ */
+ROTL : Z w -> [w] -> [w]
```
marsella, line 35: note: here and below, the `n` parameter is described as an integer in the range `[0, w)`, so I use `Z w` here to enforce that at the type level. This lets us avoid returning an option & manually checking that `n` is in range. It does require the constraint that `w > 0` since `Z 0` is not well defined, but since the spec only describes word sizes of 32 and 64 I decided this is ok.

weaversa: Another option would be to use `[width w]`.

weaversa: Some justification is that the bitvector types tend to play better with the SMT solvers.

Primitive/Keyless/Hash/SHA2/Specification.cry
weaversa: `fin w` may be redundant since `w` is bound above by `2 >= w / 32`

Primitive/Keyless/Hash/SHA2/Specification.cry
weaversa: Also, since the rotation amount is always a constant, this approach is viable:

```
ROTL : {n} (n < w) => [w] -> [w]
ROTL x = x <<< `n
```

marsella: Thanks! I like that last approach. The usage ```ROTR`{3} x``` looks more like the spec's notation $`\text{ROTR}^n(x)`$ to me, too.

Primitive/Keyless/Hash/SHA2/Specification.cry
marsella: Hm, it looks like the type checker tries to divide `w` before it realizes it has an upper bound in the last constraint, so `fin w` is needed with this formulation. But I simplified it to not divide by 32 and now it's both more readable and less redundant.

mccleeary-galois (approved):
Looks like a step in the right direction, and follows the spec cited! Thanks Marcella!

PR #115, opened 2024-08-09
Add preprocessing for SHA2

### LINKED ISSUE #110: Implement padding, parsing, and initial hash values for SHA2
Implement all the properties and functionalities from section 5 of [FIPS 180-4](https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.180-4.pdf).

- [x] Implement message padding from 5.1
- [x] Implement the sample pads as properties
- [x] Implement the message parsing in 5.2 (I think this is just calling split with the right parameters)
- [x] Set up the initial hash values in 5.3

Note that I'm punting generic SHA-512/$`t`$ to a separate issue, so skip 5.3.6 for now.

marsella:
Closes #110.

This adds the preprocessing steps from Section 5 of [FIPS 180-4](https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.180-4.pdf): padding, parsing, and initial hash values. Since the IVs are instantiation-specific, I added instantiations for the three versions of SHA2 that are named in CNSA 2.0 and Suite B: SHA256, SHA384, and SHA512.

### DIFF
--- a/Primitive/Keyless/Hash/SHA2/Instantiations/SHA256.cry
+++ b/Primitive/Keyless/Hash/SHA2/Instantiations/SHA256.cry
@@ -1,15 +1,17 @@
 /**
-Secure hash algorithm SHA256.
-
-@copyright Galois, Inc
-@author Marcella Hastings <marcella@galois.com>
-*/
+ * Secure hash algorithm SHA256.
+ *
+ * @copyright Galois, Inc
+ * @author Marcella Hastings <marcella@galois.com>
+ */
 module Primitive::Keyless::Hash::SHA2::Instantiations::SHA =
-    Primitive::Keyless::Hash::SHA2::Specification where
+    Primitive::Keyless::Hash::SHA2::Specification
+where
+    type w = 32
 
-        type w = 32
-
-        // Per [FIPS-180-4], these are the first 32 bits of the fractional
-        // parts of the square roots of the first 8 prime numbers.
-        H0 = [ 0x6a09e667, 0xbb67ae85, 0x3c6ef372, 0xa54ff53a,
-            0x510e527f, 0x9b05688c, 0x1f83d9ab, 0x5be0cd19]
+    // Per [FIPS-180-4], these are the first 32 bits of the fractional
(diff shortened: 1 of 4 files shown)


Primitive/Keyless/Hash/SHA2/Instantiations/SHA384.cry
```diff
@@ -0,0 +1,16 @@
...
+@author Marcella Hastings <marcella@galois.com>
+*/
+module Primitive::Keyless::Hash::SHA2::Instantiations::SHA384 =
+    Primitive::Keyless::Hash::SHA2::Specification where
+        type w = 64
+
+        // Per [FIPS-180-4], these are the the first sixty-four bits of the
+        // fractional parts of the square roots of the 9th - 16th prime numbers.
+        H0 = [ 0xcbbb9d5dc1059ed8, 0x629a292a367cd507,
+            0x9159015a3070dd17, 0x152fecd8f70e5939,
```
weaversa, line 14: tabs might be biting you here and elsewhere.

marsella: I think my personal spacing preferences don't align with Haskell / Cryptol sensibilities. Thanks for flagging, though. I'll check if we have internal preferences.

Primitive/Keyless/Hash/SHA2/Instantiations/SHA256.cry
```diff
@@ -0,0 +1,15 @@
+/**
```
marsella, line 1: note: this won't compile with older versions (including the standard released version) of cryptol. We had punted on #101 for the same reason, so flagging it here in case it'd be more appropriate to switch this back to normal comments for now.

Primitive/Keyless/Hash/SHA2/Instantiations/SHA384.cry
marsella: This should look nicer now!

mccleeary-galois (approved):

PR #117, opened 2024-08-16
Add SHA2 actual hashing

### LINKED ISSUE #112: Implement SHA2
We want to add a "gold standard" spec-adherent implementation of SHA2 (e.g. [FIPS 180-4](https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.180-4.pdf) Section 6).

Blocked on #109, #110. Related: #98.

- [x] Implement the hashing algorithm
- [x] Find test vectors and make sure they work
- [ ] Try to define the general properties of a hash function (idempotence, collision resistance? sensitivity?)

marsella:
Closes #112 

This adds the hash function in section 6 of [FIPS 180-4](https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.180-4.pdf).

It's not _super_ easy to follow because of the nature of functional programming and the use of list comprehensions instead of loops. I did my best to document line-by-line to make the correspondence to the spec more obvious.

Question for cryptographers: this has all the properties from the spec and it has the NIST-provided test vectors, but does not contain any more general properties of hash functions (collision resistance, small change in input => large change in output, etc.). I am not sure how to effectively express these in cryptol and I'd welcome thoughts.

Looking for review from Ryan and from one of Rawane and Brett.

### DIFF
--- a/Primitive/Keyless/Hash/SHA2/Instantiations/SHA256.cry
+++ b/Primitive/Keyless/Hash/SHA2/Instantiations/SHA256.cry
@@ -4,10 +4,12 @@
  * @copyright Galois, Inc
  * @author Marcella Hastings <marcella@galois.com>
  */
-module Primitive::Keyless::Hash::SHA2::Instantiations::SHA =
+module Primitive::Keyless::Hash::SHA2::Instantiations::SHA256 =
     Primitive::Keyless::Hash::SHA2::Specification
 where
     type w = 32
+
+    type DigestSize = 256
 
     // Per [FIPS-180-4], these are the first 32 bits of the fractional
     // parts of the square roots of the first 8 prime numbers.
--- a/Primitive/Keyless/Hash/SHA2/Instantiations/SHA256.cry
+++ b/Primitive/Keyless/Hash/SHA2/Instantiations/SHA256.cry
@@ -1,8 +1,14 @@
 /**
- * Secure hash algorithm SHA256.
+ * Secure hash algorithm SHA-256 as specified in [FIPS-180-4].
  *
  * @copyright Galois, Inc
  * @author Marcella Hastings <marcella@galois.com>
+ *
+ * [FIPS-180-4]: National Institute of Standards and Technology. Secure Hash
+ *     Standard (SHS). (Department of Commerce, Washington, D.C.), Federal
+ *     Information Processing Standards Publication (FIPS) NIST FIPS 180-4.
+ *     August 2015.
+ *     @see https://doi.org/10.6028/NIST.FIPS.180-4
  */
 module Primitive::Keyless::Hash::SHA2::Instantiations::SHA256 =
     Primitive::Keyless::Hash::SHA2::Specification
--- a/Primitive/Keyless/Hash/SHA2/Instantiations/SHA256.cry
+++ b/Primitive/Keyless/Hash/SHA2/Instantiations/SHA256.cry
@@ -1,5 +1,6 @@
 /**
- * Secure hash algorithm SHA-256 as specified in [FIPS-180-4].
+ * Instantiation of the secure hash algorithm SHA-256 as specified in
+ * [FIPS-180-4], Section 5.3.3.
  *
  * @copyright Galois, Inc
  * @author Marcella Hastings <marcella@galois.com>
--- a/Primitive/Keyless/Hash/SHA2/Instantiations/SHA384.cry
+++ b/Primitive/Keyless/Hash/SHA2/Instantiations/SHA384.cry
@@ -1,5 +1,6 @@
 /**
- * Secure hash algorithm SHA-384 as specified in [FIPS-180-4].
+ * Instantiation of the secure hash algorithm SHA-384 as specified in
+ * [FIPS-180-4], Section 5.3.4.
  *
  * @copyright Galois, Inc
  * @author Marcella Hastings <marcella@galois.com>
--- a/Primitive/Keyless/Hash/SHA2/Instantiations/SHA384.cry
+++ b/Primitive/Keyless/Hash/SHA2/Instantiations/SHA384.cry
@@ -9,6 +9,8 @@ module Primitive::Keyless::Hash::SHA2::Instantiations::SHA384 =
 where
     type w = 64
 
+    type DigestSize = 384
+
     // Per [FIPS-180-4], these are the the first sixty-four bits of the
     // fractional parts of the square roots of the 9th - 16th prime numbers.
     H0 = [
--- a/Primitive/Keyless/Hash/SHA2/Instantiations/SHA384.cry
+++ b/Primitive/Keyless/Hash/SHA2/Instantiations/SHA384.cry
@@ -1,8 +1,14 @@
 /**
- * Secure hash algorithm SHA384.
+ * Secure hash algorithm SHA-384 as specified in [FIPS-180-4].
  *
  * @copyright Galois, Inc
  * @author Marcella Hastings <marcella@galois.com>
+ *
+ * [FIPS-180-4]: National Institute of Standards and Technology. Secure Hash
+ *     Standard (SHS). (Department of Commerce, Washington, D.C.), Federal
+ *     Information Processing Standards Publication (FIPS) NIST FIPS 180-4.
+ *     August 2015.
+ *     @see https://doi.org/10.6028/NIST.FIPS.180-4
  */
 module Primitive::Keyless::Hash::SHA2::Instantiations::SHA384 =
     Primitive::Keyless::Hash::SHA2::Specification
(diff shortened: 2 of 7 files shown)


Ra1issa:
Regarding your comment on hash function properties:

Let me list some typical hash function properties in the literature to inform the discussion on how to specify and check them in cryptol. @marsella @mccleeary-galois maybe you can tell me which ones we already support?

1. **Pre-image resistance (aka. One wayness) [OW]**: Given y, it's hard to find x s.t h(x) = y
2. **Second Pre-image resistance (aka. Weak/Target Collision Resistance/) [WCR]**: Given x, it is hard to find x' != x such that h(x) = h(x')
3. **Collision Resistance (aka. Strong Collision Resistance) [CR]**: It's hard to find any two x and x' (x != x') such that h(x) = h(x') 
4. **Non-Malleability**: Given h(x) it's hard to find h(x') where R(x, x') for some class of relations (the property holds for that class) 
5. **Pseudorandomness**: h is indistinguishable for a random oracle. 

When we talk about SHA2, we are interested in the first 3 properties ([see](https://csrc.nist.gov/projects/hash-functions) this NIST page). Here is how they are related to one another ([ref](https://www.cs.ucdavis.edu/~rogaway/papers/relates.pdf)):

- CR => WCR
- CR does **not** => PR
- PR does **not** => WCR
- PR does **not** => CR

In light of this, can you explain your reasoning in the sentence below? I am not sure which of the properties of secure hashing implies it. 
 > small change in input => large change in output

marsella:
> In light of this, can you explain your reasoning in the sentence below? I am not sure which of the properties of secure hashing implies it.
> > small change in input => large change in output

Hm, I guess this idea of sensitivity isn't formal. Maybe it's related to one of the notions of collision resistance, where you could say that if small change in input produced a small change in output, an adversary would be more likely to find a collision by tweaking plaintexts.

I pulled this out of [the properties listed](https://htmlpreview.github.io/?https://github.com/GaloisInc/hacrypto/blob/master/arch/index.html#class_chart:ANY) from the [HACrypto repo](https://github.com/GaloisInc/hacrypto), but didn't think about it beyond that. I think it's reasonable to focus on formalizing the properties you listed rather than these ones.

I'm not sure how to encode "hard to find" properties in Cryptol. I think I gravitated toward sensitivity because it felt more measurable, like I could, say, count the number of different bits in the digest to compute how different two outputs are. I'll ask internally about paradigms to approach some of these.

mccleeary-galois (approved):
In terms of following the spec as we want this LGTM. I think we should try and push the boundaries of the properties we are trying to prove here as well as that would be good to show in here. 

One thing to note is that I was getting a `No fences` on `:check-docstrings` but I couldn't find the property it couldn't find the fence on?

```
Cryptol> :load Primitive/Keyless/Hash/SHA2/Tests/SHA256.cry 
Loading module Cryptol
Loading module `where` argument of Primitive::Keyless::Hash::SHA2::Instantiations::SHA256
Loading interface module `parameter` interface of Primitive::Keyless::Hash::SHA2::Specification
Loading module Primitive::Keyless::Hash::SHA2::Specification
Loading module Primitive::Keyless::Hash::SHA2::Instantiations::SHA256
Loading module Main
Main> :check-docstrings


Checking Main::abcWorks

:prove abcWorks
Q.E.D.
(Total Elapsed Time: 0.045s, using "Z3")


Checking Main::emptyStringWorks

:prove emptyStringWorks
Q.E.D.
(Total Elapsed Time: 0.033s, using "Z3")


Checking Main::alphabet448

:prove alphabet448
Q.E.D.
(Total Elapsed Time: 0.065s, using "Z3")


Checking Main::alphabet896

:prove alphabet896
Q.E.D.
(Total Elapsed Time: 0.060s, using "Z3")

Successes: 4, No fences: 1, Failures: 0
```

marsella:
> One thing to note is that I was getting a No fences on :check-docstrings but I couldn't find the property it couldn't find the fence on? 

I put a docstring at top level on the module, could that be it? @glguy

marsella:
I might make a separate issue to dive deeper on properties of hash functions. Would be related to #98 -- if we write a hash interface then we could define all the properties over hash functions generically and then instantiate them for each concrete hash that we care about.

glguy:
> > One thing to note is that I was getting a No fences on :check-docstrings but I couldn't find the property it couldn't find the fence on?
> 
> I put a docstring at top level on the module, could that be it? @glguy

Yes, module headers are a place that can have docstrings and fences

Primitive/Keyless/Hash/SHA2/Instantiations/SHA256.cry
```diff
@@ -1,8 +1,14 @@
 /**
- * Secure hash algorithm SHA256.
+ * Secure hash algorithm SHA-256 as specified in [FIPS-180-4].
```
Ra1issa, line 2: nitpick: I would call out that this is just the instantiation of the algorithm and refer to the specific section in the FIPS,so:

***Instantiation** of Secure hash algorithm SHA-256 as specified in [FIPS-180-4] **section 5.3.3***.

Primitive/Keyless/Hash/SHA2/Instantiations/SHA384.cry
```diff
@@ -1,8 +1,14 @@
 /**
- * Secure hash algorithm SHA384.
+ * Secure hash algorithm SHA-384 as specified in [FIPS-180-4].
```
Ra1issa, line 2: nitpick: Same as above

_Instantiation of Secure hash algorithm SHA-384 as specified in [FIPS-180-4] section 5.3.4._

Primitive/Keyless/Hash/SHA2/Instantiations/SHA512.cry
```diff
@@ -1,8 +1,14 @@
 /**
- * Secure hash algorithm SHA512.
+ * Secure hash algorithm SHA-512 as specified in [FIPS-180-4].
```
Ra1issa, line 2: nitpick: Same as above

_Instantiation of Secure hash algorithm SHA-512 as specified in [FIPS-180-4] section 5.3.5._

Ra1issa:
**Overall**: This PR looks good to me. I added a few nitpicks. I was able to understand what each of the modified files was set out to do and was able to follow along with the FIPS doc. This is an overall code quality improvement over the existing code 👍 

**Question/Request**: The one place that was harder for me to follow was in the Specification.cry file. This is possibly because of my inexperience with Cryptol and you should feel free to ignore this comment if my concerns do not represent typical Cryptol users. Could you maybe explain to me what things defined in `private` are set out to do ? Is this where some of the properties of SHA2 are specified?  

**Discussion**: One more hash function property that I forgot to add to the list is that the hash function has to be **deterministic** in x, meaning that every time you apply h to x you get the same output y (i.e. its not random).

Primitive/Keyless/Hash/SHA2/Specification.cry
```diff
@@ -24,13 +24,20 @@ module Primitive::Keyless::Hash::SHA2::Specification where
 
 parameter
     /**
-     * Word size.
+     * Length of words, in bits, that are used during hashing.
+     *
      * The specification defines words to be either 32 or 64 bits.
```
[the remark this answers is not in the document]
marsella: Yes, good call! I added some notes on which legal SHAs get which word lengths.

marsella:
> what things defined in private are set out to do ? Is this where some of the properties of SHA2 are specified?

The `private` label is less about the spec itself and more about how I expect other specs to interact with this module. My working assumption is that anyone who wants to use a hash function will just call `hash`, and will not need to call the component functions (`pad`, `parse`, any of the linear functions `Ch`/`Sigma` etc). They will likely also need to know the `MaxMessageWidth` so they can use it as a type constraint on their own functions. So `hash` and `MaxMessageWidth` are public (by default), and everything else is private.

I don't have a particularly principled reason for putting the properties in private. I think the current plan for checking properties is to load files and use the `:check-docstrings` command, which can interact with private properties. Mostly I just wanted those properties to live close to where they are in the spec, so they ended up private because I tabbed everything at once.

Ra1issa:
@marsella would it make sense to move everything under `private` till the end of the file, and move the doc for SHA2's pseudocode right before it so that it lives closer to the parameters ?

Ra1issa (approved):

marsella:
> would it make sense to move everything under private till the end of the file, and move the doc for SHA2's pseudocode right before it so that it lives closer to the parameters ?

I considered that. I decided ultimately that I wanted to maintain the order of functions from the spec more than I wanted the file to be naturally ordered to someone who was looking at it independently. Maybe @mccleeary-galois can chime in on whether that was a reasonable choice; it's worth deciding now because I'll be repeating whatever pattern we like for many more specs.

mccleeary-galois:
> > would it make sense to move everything under private till the end of the file, and move the doc for SHA2's pseudocode right before it so that it lives closer to the parameters ?
> 
> I considered that. I decided ultimately that I wanted to maintain the order of functions from the spec more than I wanted the file to be naturally ordered to someone who was looking at it independently. Maybe @mccleeary-galois can chime in on whether that was a reasonable choice; it's worth deciding now because I'll be repeating whatever pattern we like for many more specs.

I could be convinced to switch but I prefer the current set up as is as it made it easy for me to follow along with the spec in hand.

PR #119, opened 2024-08-21
Remove old EC / ECDSA implementations

### LINKED ISSUE #113: Remove duplicated versions of ECDSA and ECs
At time of writing there are 4 implementations of the NIST prime field elliptic curves and 3 implementations of ECDSA. This is redundant and we should remove the ones that don't meet our gold standards.

Right now the closest to gold standard are the versions added in #99 and #96. We should carefully review the other versions to see if there are useful things contained in them. If possible, we should port those things to the standard versions. If not, document what we want to keep in this issue.

- [x] Review Z-type ECDSA + curve implementations
- [x] Review integer version of ECDSA
- [x] Review the one-file standalone version of EC
- [x] Review the other EC version
- [x] Update the EC and ECDSA READMEs to reflect which versions were deleted.

marsella:
Closes #113.

Updates the READMEs to give more context on history and point towards other versions of ECDSA and EC living elsewhere. But mostly deletes.


Common/EC/README.md
```diff
@@ -2,40 +2,12 @@
...
-The second is optimized for SAW proofs against a Java implementation. It has some generic components but is only fully instantiated for the P-384 curve. It has a corresponding ECDSA implementation in `Primitive/Asymmetric/Signature/ecdsa.cry`. It does not instantiate the elliptic curve interface.
-```
-Common/EC/
-+ ec_point_ops.cry
-+ p384_ec_mul.cry
-+ p384_ec_point_ops.cry
-+ p384_field.cry
-+ ref_ec_mul.cry
-```
+This is one of many, many implementations of the NIST prime field curves that have been written in Cryptol over the years. One grandparent of particular interest is a 2011 curve implementation that was used to verify a Java implementation of ECDSA in combination with [SAWScript](https://saw.galois.com/). This implementation no longer lives in this repo, but it can be seen together with the Java code and the SAW scripts used in the full verification toolchain [in the examples directory of the SAWScript repository](https://github.com/GaloisInc/saw-script/tree/master/examples/ecdsa).
```
marsella, line 9: Two links questions:
- Should I link to the SAWScript website or its github page? I think the website is a bit out of date but it has a good quick answer to "what is saw??"
- Should I pin to a specific commit when pointing to the saw example or is master OK? In particular, do we expect to rearrange the structure of that repo such that this link will break? cc: @sauclovian-g

mccleeary-galois: Let's link to the github page for now.

sauclovian-g: There will almost certainly be a general reorg of the saw-script tree eventually... but I don't see that extending to moving the ```examples``` dir, so you're probably safe to point to it on master.

marsella: Thanks! I updated the saw link.

mccleeary-galois (approved):
Thanks for cleaning this up Marcella!

PR #124, opened 2024-08-26
CBC mode refactor/improvements

staslyakhov:
Addresses, but does not complete, #80 : CBC refactor and generalization.

Follow the structure established in #83 for CTR, and applies it to CBC mode.

* CBC implementation independent of cipher used (via `CipherInterface`)
* Instantiation for all 3 AES key sizes
* Test vectors for all 3 key sizes (encryption + decryption) taken from NIST-SP-800-38A
* Improved documentation


Primitive/Symmetric/Cipher/Block/Modes/CBC.cry
```diff
@@ -1,21 +1,60 @@
-// Cryptol CBC Implementation
-// Copyright (c) 2010-2018, Galois Inc.
-// www.cryptol.net
-// Author: Ajay Kumar Eeralla
+/*
+ * Cipher Block Chaining mode of operation, as defined in [NIST-SP-800-38A], Section 6.2.
+ *
+ * ⚠️ Warning ⚠️: CBC mode requires that the initialization vector (IV) is generated "unpredictably".
```
staslyakhov, line 4: CTR has a warning about not duplicating counters, so I figured I'd leave a warning about having unpredictable IVs here. Is that a good place to document such requirements?

Primitive/Symmetric/Cipher/Block/Modes/CBC.cry
```diff
@@ -1,21 +1,60 @@
...
-    where cs = [ enc k (p ^ c') | p <- ps | c' <- [iv] # cs ]
+/**
+ * CBC encryption: [NIST-SP-800-38A] Section 6.2.
+ *
+ * Parameters: key, initialization vector, plaintext
+ *
+ * ⚠️ Warning ⚠️: CBC mode requires that the initialization vector (IV) is generated "unpredictably".
+ * See Appendix C of [NIST-SP-800-38A] for discussion.
+ */
+encrypt : {n} (fin n) => [C::KeySize] -> [C::BlockSize] -> [n][C::BlockSize] -> [n][C::BlockSize]
```
staslyakhov, line 32: I didn't really change the implementation of encrypt/decrypt much. Does it look close enough to the spec? Is this something we frequently worry about when writing cryptol?

For reference, here's the definition from NIST-SP-800-38A:

image

Primitive/Symmetric/Cipher/Block/Tests/TestAES_CBC.cry
```diff
@@ -0,0 +1,139 @@
...
+import Primitive::Symmetric::Cipher::Block::Instantiations::AES128_CBC as AES128_CBC
+import Primitive::Symmetric::Cipher::Block::Instantiations::AES192_CBC as AES192_CBC
+import Primitive::Symmetric::Cipher::Block::Instantiations::AES256_CBC as AES256_CBC
+
+/**
+ * ```repl
+ * :prove aes128_cbc_encrypt_vector
+ * ```
+ */
+property aes128_cbc_encrypt_vector = (AES128_CBC::encrypt k iv plaintext) == ciphertext
```
marsella, line 26: thought: it looks like the encrypt and decrypt vectors use the same inputs. I'm torn between reducing that duplication by testing encrypt and decrypt in a single test vector vs having the more precise tests that you have now. 

No action needed, just a thought.

Primitive/Symmetric/Cipher/Block/Modes/CBC.cry
staslyakhov: Yeah I spent a non-trivial amount of time thinking about how to emulate $C_{j - 1}$. I feel as though this will come up a lot so that would be nice to standardize. Maybe there's a discussion to be had about making cryptol support variable names with weird subscripts somehow?

Primitive/Symmetric/Cipher/Block/Tests/TestAES_CBC.cry
staslyakhov: I didn't even notice! The spec doesn't make it too obvious. I'm conflicted too: on the one hand it's nice to know what exactly is failing (is it encrypt or is it decrypt?), but on the other hand that is quite a bit of duplication.

I think currently I'm leaning toward keeping it as is since it's more obvious that it follows the spec, which also duplicates the test vectors. I noticed CTR seems to do it in the same way, so if we want to do it differently we should probably decide how and keep it consistent across the specs.

marsella: Oh yeah, if that's how the spec describes them then let's keep them as-is.

Primitive/Symmetric/Cipher/Block/Modes/CBC.cry
```diff
@@ -1,21 +1,60 @@
...
 
-cbcEnc : {n, k} (fin n, fin k) => ([k]->block -> block) -> [k] -> iv -> [n]block -> [n]block
-cbcEnc enc k iv ps = cs
-    where cs = [ enc k (p ^ c') | p <- ps | c' <- [iv] # cs ]
+/**
+ * CBC encryption: [NIST-SP-800-38A] Section 6.2.
+ *
+ * Parameters: key, initialization vector, plaintext
+ *
+ * ⚠️ Warning ⚠️: CBC mode requires that the initialization vector (IV) is generated "unpredictably".
```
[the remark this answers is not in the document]
staslyakhov: I liked your wording! Added that and mentioned Appendix D.

Primitive/Symmetric/Cipher/Block/Modes/CBC.cry
staslyakhov: I made an attempt to make the variables more similar. Let me know what you think! Then I can make an issue (or maybe extend the existing issue?) to do the same for CTR.

marsella: I like it! Thanks for updating.

I looked through some recent PRs to see how I've handled this elsewhere. The closest thing is in SHA-2, where I [skipped the `-1` entirely](https://github.com/GaloisInc/cryptol-specs/pull/117/files#diff-828fceee934f75ae76693d07d677770796c9d9053347a993feefa74378ea2695R327) and documented the difference.

marsella (approved):
Looks good, thank you!

marsella:
Oh, one final request: please update #80 to
- separate the checkbox for CBC / CFB parameterization into two and mark off the one you did
- add a checkbox for fixing up naming in CTR

staslyakhov:
> Looks good, thank you!

Thanks for the review! I updated the issue to separate the tasks for different modes and mark the CBC refactor as completed.

PR #125, opened 2024-08-27
Add ECDH primitives

### LINKED ISSUE #123: Add ECDH primitive
We want to support key agreement via ECDH, as specified in [NIST SP 800-56A](https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-56Ar3.pdf). As a first step, we need to support the basic ECDH primitive specified in there, as well as some validation for the keys involved.

- [x] Define a type alias for public and private keys
- [x] Add validation functions for public and private keys
- [x] Define the field-to-byte conversion routine or make sure it matches cryptol defaults.
- [x] Add ECC CDH primitive from 5.7.1.2 with warnings about use (validation required on inputs, key derivation required before using output; destroying intermediate reprs necessary)
- [x] Write a follow-up issue to add test vectors and make sure the implementation aligns with CAVP and KASVS requirements - see #136

marsella:
Closes #123.

This adds the ECC_CDH primitive (probably what most people would call ECDH) from [NIST SP 800-56A rev 3](https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-56Ar3.pdf). The specification is very long and includes a lot of things that are hard to instantiate in cryptol, like key validation routines that involve trusted third parties and key generation based on sampling randomness.

This PR adds the most absolute barebones primitive needed to compute a shared secret. On its own this is _totally inadequate_ for any actual key generation: it's missing parameter validation, key derivation, key confirmation, and any notion of (input) key management. I added some docs to that effect, but I would like input on whether it's clear that this is not secure as-is and needs a bunch of additional routines to approach security.

There are also a fair number of changes to the EC interface. The most complex one is the `xCoord` method, since that required propagating some changes to ECDSA. The rest are all straightforward additions.

I added test vectors for P-256 to make sure it works. I will write up a separate issue to add other instantiations and their corresponding test vectors. 

Crypto review request (looking for one crypto review from either @b13decker or @Ra1issa):
- Did I miss any insecurities of this standalone primitive?
- Are there any properties I should write up for the key validation methods?

General review request:
- Are you alarmed by the documentation?
- Are there other basic primitives that would be useful to include in this first pass?
- Are there other instantiations for specific curves that you would like to see?


Primitive/Asymmetric/Generation/ECDH/Tests/ECDH_P256.cry
```diff
@@ -0,0 +1,344 @@
+/*
+ * Run the CAVP test vectors.
+ */
```
marsella, line 3: todo: add reference to CAVP vectors. Add copyright notices.

Primitive/Asymmetric/Generation/ECDH/Specification.cry
```diff
@@ -0,0 +1,161 @@
...
+ * Get assurance of private-key validity.
+ *
+ * A valid private key is in the interval `[1, n-1]`, where `n` is the order
+ * of the base point for the curve.
+ * The upper bound is assured by the `SigningKey` type itself.
+ *
+ * [SP-800-56Ar3] Section 5.6.2.1.2.
+ */
+signingKeyValid : SigningKey -> Bool
+signingKeyValid d = d != 0
```
mccleeary-galois, line 83: Doesn't this need to be a lower bound and not just an inequality? Or am I confused by the `Z` type in SigningKey?

Primitive/Asymmetric/Generation/ECDH/Tests/ECDH_P256.cry
```diff
@@ -0,0 +1,362 @@
...
+ * "IUT" stands for "implementation under test".
+ */
+cavpTest QCAVSx QCAVSy dIUT QIUTx QIUTy Z = zMatches && iutKeysValid && cavsKeyValid where
+    Q_C = P256::Affine (BVtoZ QCAVSx) (BVtoZ QCAVSy)
+    Q_I = P256::Affine (BVtoZ QIUTx) (BVtoZ QIUTy)
+    d_I = BVtoZ dIUT
+    zMatches = case ECDH::ECC_CDH d_I Q_C of
+        Some actualZ -> actualZ == (split Z)
+        None -> False
+    iutKeysValid = ECDH::signingKeyValid d_I
```
mccleeary-galois, line 34: This is more food for thought than an ask... As of right now we have been adding the test vectors outlined in the spec and I liked the way you set up this test file, but should we also add some tests for expected invalid cases?

mccleeary-galois (approved):
LGTM overall except for some clarity questions. I was able to follow along with the spec cited and it seemed all kosher to me.

Answered questions posed below.

Are you alarmed by the documentation?

- I think it is important that we call out the current issues like you have done so.

Are there other basic primitives that would be useful to include in this first pass?

- None that I could think of, I think more will come to light as we add more curves and instantiations.

Are there other instantiations for specific curves that you would like to see?

- I don't think anything should block getting in this PR but lets chat about prioritization here soon.

Primitive/Asymmetric/Generation/ECDH/Specification.cry
```diff
@@ -1,4 +1,12 @@
 /**
+ * Specification for some of the cryptographic primitives used in pair-wise key
+ * establishment as defined in [SP-800-56Ar3].
+ *
+ * This does not implement any complete key establishment schemes! It
```
Ra1issa, line 5: Re comment above

Primitive/Asymmetric/Generation/ECDH/Specification.cry
```diff
@@ -110,3 +120,42 @@ verifyingKeyPartiallyValid Q = EC::isValid Q && ~(EC::isInfinity Q)
...
+ * Compute a shared secret `Z` using the domain parameters, the other party's
+ * public key (`QB`) and one's own private key (`dA`).
+ *
+ * This deviates from the spec in one important way: All intermediate values,
+ * including `P` and `z`, should be destroyed (zeroized) before providing
+ * output. Cryptol cannot express this; implementors must manually verify that
+ * all potentially sensitive local data is destroyed.
+ * [SP-800-56Ar3] Section 5.7.1.2.
+ *
+ * ⚠ Usage notes ⚠: This routine is not secure if used in isolation! It should be
```
Ra1issa, line 134: Yep this is great. I would all caps "WARNING: USAGE NOTES"

Primitive/Asymmetric/Generation/ECDH/Specification.cry
```diff
@@ -110,3 +120,42 @@ verifyingKeyPartiallyValid Q = EC::isValid Q && ~(EC::isInfinity Q)
 keyPairIsConsistent : SigningKey -> VerifyingKey -> Bool
 keyPairIsConsistent d Q = EC::pointEq expectedQ Q where
     expectedQ = EC::scmul (fromZ d) EC::G
+
+/**
+ * Compute a shared secret `Z` using the domain parameters, the other party's
+ * public key (`QB`) and one's own private key (`dA`).
+ *
+ * This deviates from the spec in one important way: All intermediate values,
```
Ra1issa, line 128: Needs a warning header like what I've described before

Primitive/Asymmetric/Generation/ECDH/Specification.cry
```diff
@@ -110,3 +120,42 @@ verifyingKeyPartiallyValid Q = EC::isValid Q && ~(EC::isInfinity Q)
...
+ * output. Cryptol cannot express this; implementors must manually verify that
+ * all potentially sensitive local data is destroyed.
+ * [SP-800-56Ar3] Section 5.7.1.2.
+ *
+ * ⚠ Usage notes ⚠: This routine is not secure if used in isolation! It should be
+ * used as a component of one of the key agreement schemes described in
+ * [SP-800-56Ar3] Section 6. These schemes handle (input) key management and
+ * validation, (output) key derivation, and optional key confirmation.
+ *
+ * - Failure to use a valid key agreement scheme can compromise the security of
```
Ra1issa, line 139: Re emphasizing warning

Primitive/Asymmetric/Generation/ECDH/Specification.cry
```diff
@@ -110,3 +120,42 @@ verifyingKeyPartiallyValid Q = EC::isValid Q && ~(EC::isInfinity Q)
...
+ * the `SigningKey`s used in `ECC_CDH`! This routine does not validate the
+ * other party's public key. A malicious party can claim invalid curve points
+ * for her public key `QB` and collect residues to derive the honest party's
+ * private key!
+ *
+ * - The output of this method is a _shared secret_, not a key!
```
Ra1issa, line 145: Re emphasizing warning

Primitive/Asymmetric/Generation/ECDH/Specification.cry
marsella: The `Z` type will prevent negative numbers. It represents the integers mod `n` so essentially the range `[0, n-1]`. The only thing in that range we need to disallow is 0.

Technically you can instantiate Z-types from a negative number, but it'll pull it into the mod group:
```
> -1 : Z 13
12
```
so, for the purposes of this function, we won't ever get a negative number because it'll be "canonicalized" by definition into something in the appropriate range. 

This implementation doesn't address the question of where signing keys come from. We might get a representational problem if someone is improperly generating signing keys and thinks their secret is a negative number, but it'll still be equivalent in `Z n` so I'm not sure if it'll be a problem mathematically.

Ra1issa (commented):
Here is a first batch of comments. I will scrutinize the changes against the spec in a bit. Overall, the changes you've made vastly improve readability and clearly outlines the limitations of this work

Ra1issa:
Can someone clarify to me why this exists and what its intended use is if "On its own this is totally inadequate for any actual key generation" ? Its not just a couple of minor things that are missing for this to be a valid and verified key establishment, its core properties that @marsella outlined both in her comments and the docs.

Primitive/Asymmetric/Generation/ECDH/Specification.cry
Ra1issa: @mccleeary-galois is it worth adding the explanation that @marsella just gave to the docs? I think that you would have a better feel for what the intended audience know/does not know.

mccleeary-galois: Nope, I am good with this. This was me mostly having an issue context switching and was in a domain where Z meant the Integers rather than Modular integers which is documented in Cryptol.

Ra1issa:
It would also be super helpful to have a single place that highlights all the parts of [NIST SP 800-56A rev 3](https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-56Ar3.pdf) that are validated and those that are not. Maybe this already exists and I just missed it ?

Primitive/Asymmetric/Generation/ECDH/Tests/ECDH_P256.cry
marsella: Yeah, I think that would be very reasonable to avoid dumb false positives (e.g. imagine a broken verification function that just returns true for everything). I made #136.

marsella:
Thanks @Ra1issa! I updated the docs in a few ways:
- Added a more obvious warning at the top that this isn't key establishment!
- Added a list at the top of all the things we don't include here (hopefully this addresses your last comment to a certain degree) and a short explanation of why each one is a necessary part of key establishment
- Rephrased the warnings in the per-function docs and made sure they were all labeled appropriately.

Hopefully with all of these things combined it will be very hard for someone to accidentally misuse this. There is significantly more warning text than code in the file now.

Primitive/Asymmetric/Generation/ECDH/Specification.cry
```diff
@@ -0,0 +1,209 @@
...
+ * - It must have the expected representation for an element in the underlying
+ *   field;
+ * - It must be a point on the correct curve; and
+ * - It must have the correct order.
+ *
+ * [SP-800-56Ar3] Section 5.6.2.3.3.
+ */
+verifyingKeyFullyValid : VerifyingKey -> Bool
+verifyingKeyFullyValid Q = verifyingKeyPartiallyValid Q && correctOrder where
+    correctOrder = EC::isInfinity (EC::scmul `EC::n Q)
```
staslyakhov, line 142: This might not matter here but I figured I'd point this out just in case: if $g^n = 1$ (where $g$ is some base element and 1 is identity), it's not always true that $g$ must have order $n$.

There might some other $k < n$ such that $g^k$ also equals 1, in which case `k` would be the order (assuming there isn't some number less than $k$ where that's true).

marsella: In this case we're using NIST's standard definition of domain parameters for elliptic curve groups, where the base point `G` generates a large subgroup of prime order `n`. Note that the order of the base point `n` is not necessarily the same as the modulus for the field over which the curve is defined (e.g. your curve points are in $GF(q)$ but your large interesting subgroup is order `n < q`). 

By definition, any correctly generated verifying key `Q = dG` for the secret `d` and the curve base point `G` and therefore should have order `n` because `G` has order `n`.

If you use this with invalid domain parameters then the situation you describe might be true. I think the assumption this makes is that the domain parameters are valid, therefore if `nQ = 1` and it's on the curve then it's in the subgroup (because `n` is prime so there is no `k|n` -- or could we have a `k < n` that doesn't divide it but still multiplies to the identity?) and we're happy.

staslyakhov: I think the assumption about `n` being prime is exactly the thing that makes it all work :+1: . 

This is off topic so feel free to ignore, but what you said made me curious: does NIST actually require that `n < q` or were you just using that as an example to illustrate that they're not the same thing? It should be possible for the elliptic curve field to have a larger cardinality than the field over which it is defined (of course one could avoid such curves).

marsella: Huh, I guess I filled that in for myself based on the curves I was using. I've been referencing [NIST SP 800-186](https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-186.pdf) Section 2 and [186-5](https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.186-5.pdf) Section 6 as my baseline and neither of them explicitly say `n < q`.

staslyakhov: Gotcha. Thanks for the discussion--it was really helpful!

Issue #132, opened 2024-08-30
 Bring SHA3 implementation up to gold standard


marsella:
We have a pretty good version of SHA3 (both the hash functions and the XOFs), but it doesn't quite meet [our gold standard](). This issue lists some initial action items with one broad follow-up task. It may be broken into several PRs if there are natural stopping points along the way.

- [x] Add documentation across the module to map back to spec
- [x] Explain representation differences & how we access state
- [x] Search for examples in the spec that could be encoded as properties
- [x] Add docstrings to the properties
- [x] Add separate tests for each instantiation of SHA3 and add docstrings
- [x] Add tests for SHAKEs with docstrings
- [x] Address any other concerns that arise while fixing the above.

PR #137, opened 2024-09-06
Update structure of SHA3

marsella:
Addresses part of #132, but doesn't complete it.

This PR updates the high-level structure of SHA-3 to more closely imitate the spec. In particular, it:
- adds parameters to the `keccak` implementation to reduce per-function parameterization
- rearranges code to more closely match the order that things appear in the spec and to make many methods private.
- adds documentation on the high-level functions

~One effect of this is to change the way that the hash and XOF functions are called; a remaining TODO for this PR is to fix all of the downstream failures that result from this API change.~ Addressed this with clarified types instead of changing a dozen downstream files! Yay!

Several things here are being left to future PRs. I'll write up issues for these before this PR is merged.
- many functions don't really look like their spec equivalents. Lower-level modifications of this kind are being saved for a separate PR
- some of the keccak implementation is overly specific to the SHA-3 instantiation parameters, when the spec writes it as more generic. Reducing hard-coded parameters (`nr` specifically) that don't affect the SHA-3 instantiations is being saved for a separate PR.

### DIFF
--- a/Primitive/Keyless/Hash/SHA3/SHA3.cry
+++ b/Primitive/Keyless/Hash/SHA3/SHA3.cry
@@ -1,3 +1,16 @@
+/*
+ * SHA-3 hash function, as specified in [FIPS-202].
+ *
+ * [FIPS-202]: National Institute of Standards and Technology. SHA-3 Standard:
+ *     Permutation-Based Hash and Extendable-Output Functions. (Department of
+ *     Commerce, Washington, D.C.), Federal Information Processing Standards
+ *     Publication (FIPS) NIST FIPS 202. August 2015.
+ *     @see https://dx.doi.org/10.6028/NIST.FIPS.202
+ *
+ * @copyright Galois, Inc.
+ * @author Ajay Kumar Eeralla
+ * @editor Marcella Hastings <marcella@galois.com>
+ */
 module Primitive::Keyless::Hash::SHA3::SHA3 where
 import Primitive::Keyless::Hash::keccak where
     type b = 1600
@@ -8,5 +21,12 @@ parameter
   type digest : #
   type constraint (fin digest, digest >= 224, digest <= 512)
 
+/**
+ * SHA-3 hash function specification.
+ * [FIPS-202] Section 6.1.
+ *
+ * Note that the specification of `c` is above, in the instantiation of the
+ * `keccak` module.
+ */
 sha3 : {n} (fin n) => [n] -> [digest]
 sha3 M = Keccak`{d=digest} (M # 0b01)
--- a/Primitive/Keyless/Hash/SHA3/SHA3.cry
+++ b/Primitive/Keyless/Hash/SHA3/SHA3.cry
@@ -1,12 +1,12 @@
 module Primitive::Keyless::Hash::SHA3::SHA3 where
-import Primitive::Keyless::Hash::keccak
+import Primitive::Keyless::Hash::keccak where
+    type b = 1600
+    type nr = 24
 
 parameter
   type digest : #
   type constraint (fin digest, digest >= 224, digest <= 512)
 
-type total = 1600
-
 sha3 : {n} (fin n) => [n] -> [digest]
 sha3 M =
-  take`{digest} (Keccak `{r = total - 2 * digest, c = 2 * digest} (M # 0b01))
+  take`{digest} (Keccak `{r=1600-(2*digest)} (M # 0b01))
--- a/Primitive/Keyless/Hash/SHA3/SHA3.cry
+++ b/Primitive/Keyless/Hash/SHA3/SHA3.cry
@@ -2,11 +2,11 @@ module Primitive::Keyless::Hash::SHA3::SHA3 where
 import Primitive::Keyless::Hash::keccak where
     type b = 1600
     type nr = 24
+    type c = 2 * digest
 
 parameter
   type digest : #
   type constraint (fin digest, digest >= 224, digest <= 512)
 
 sha3 : {n} (fin n) => [n] -> [digest]
-sha3 M =
-  take`{digest} (Keccak `{r=1600-(2*digest)} (M # 0b01))
+sha3 M = Keccak`{d=digest} (M # 0b01)
--- a/Primitive/Keyless/Hash/SHAKE/SHAKE128.cry
+++ b/Primitive/Keyless/Hash/SHAKE/SHAKE128.cry
@@ -1,11 +1,12 @@
 module Primitive::Keyless::Hash::SHAKE::SHAKE128 where
-import Primitive::Keyless::Hash::keccak
+import Primitive::Keyless::Hash::keccak where
+    type b = 1600
+    type nr = 24
 import Primitive::Keyless::Hash::utils (toBytes)
 
-type total = 1600
 type secBits = 128
 
-shake128 M = Keccak `{r = total - 2 * secBits, c = 2 * secBits } (M # 0b1111)
+shake128 M = Keccak `{r = 1600 - 2 * secBits } (M # 0b1111)
 
 property k1 = join (toBytes (take`{512} (shake128 []))) == join [ 0x7f9c2ba4e88f827d616045507605853e
                                                              , 0xd73b8093f6efbc88eb1a6eacfa66ef26
--- a/Primitive/Keyless/Hash/SHAKE/SHAKE128.cry
+++ b/Primitive/Keyless/Hash/SHAKE/SHAKE128.cry
@@ -26,10 +26,12 @@ import Primitive::Keyless::Hash::utils (toBytes)
  * SHAKE128 extendable-output function.
  * [FIPS-202] Section 6.2.
  *
+ * This supports any output length `d`, including infinite length.
+ *
  * Note that the specification of `c` is above, in the instantiation of the
  * `keccak` module.
  */
-shake128 : {m, d} (fin m, fin d) => [m] -> [d]
+shake128 : {d, m} (fin m) => [m] -> [d]
 shake128 M = Keccak (M # 0b1111)
 
 /**
--- a/Primitive/Keyless/Hash/SHAKE/SHAKE128.cry
+++ b/Primitive/Keyless/Hash/SHAKE/SHAKE128.cry
@@ -35,12 +35,12 @@ shake128 : {d, m} (fin m) => [m] -> [d]
 shake128 M = Keccak (M # 0b1111)
 
 /**
- * Example showing how to call shake128 using an explicit type parameter.
+ * Example showing how to call shake128 using a type parameter.
  * ```repl
  * :prove k1
  * ```
  */
-property k1 = join (toBytes (shake128`{d=512} [])) == expected_result where
+property k1 = join (toBytes (shake128`{512} [])) == expected_result where
     expected_result = join [
         0x7f9c2ba4e88f827d616045507605853e,
         0xd73b8093f6efbc88eb1a6eacfa66ef26,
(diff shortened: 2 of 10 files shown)


weaversa:
It kind of feels like something is wrong the with `sha3` if `join (toBytes ...` is needed every time its run. I've struggled quite a bit with SHA3's bit reorganization routine and haven't yet found an elegant way to avoid asking the user to rejigger values at the function boundary. If you consider updates here, please keep in in the loop.

Primitive/Keyless/Hash/keccak.cry
```diff
@@ -1,119 +1,299 @@
...
+     * [FIPS-202] Section 3.1.
+     */
+    type State = [5][5][w]
+
+    /**
+     * Convert a string into a state array.
+     * [FIPS-202] Section 3.1.2.
+     */
+    unflatten : [b] -> State
+    unflatten p = transpose (groupBy`{5} (reverse (groupBy`{w} (reverse p))))
```
marsella, line 112: note: For this PR I focused on cleaning up type signatures and documentation. The actual implementation of the following functions does not look very close to the spec; I have to write a follow-up issue to address fixing that up. Applies to `unflatten`, `flatten`, `θ`, `ι`, `χ`, `π`, `ρ`, `RC` and `RCs`.

I recommend reviewers don't spend time right now looking through these implementations right now.

Primitive/Keyless/Hash/keccak.cry
```diff
@@ -1,119 +1,299 @@
...
+     *      (m + 2) /^ x      : The multiplier of `x` that determines the
+     *                        : ultimate length of the message + padding
+     * x * ((m + 2) /^ x)     : The total length of the message + padding
+     * x * ((m + 2) /^ x) - m : The total length of the padding alone
+     */
+    pad : {x, m} (fin x, fin m, x >= 1) => [x * ((m + 2) /^ x) - m]
+    pad = [1] # zero # [1]
+
+    /**
+     * Keccak-f family of permutations.
```
marsella, line 237: note: this version doesn't have `Keccak-p` at all; that's the fully generic version of `Keccak` that allows arbitrary number of rounds. It's needed to match the spec properly but _not_ needed to instantiate the SHA3 hash functions or SHAKE XOFs, so I'm going to punt it to a separate PR.

Primitive/Keyless/Hash/keccak.cry
marsella, line 1: note: I'm sorry about this diff! I changed both the spacing and the order of things in here, so it's hard to trace between the old and new versions. I recommend focusing on documentation, ordering with respect to the spec, and presence / absence of all the relevant functions.

Primitive/Keyless/Hash/SHAKE/SHAKE128.cry
```diff
@@ -1,34 +1,92 @@
...
+/**
+ * SHAKE128 extendable-output function.
+ * [FIPS-202] Section 6.2.
+ *
+ * This supports any output length `d`, including infinite length.
+ *
+ * Note that the specification of `c` is above, in the instantiation of the
+ * `keccak` module.
+ */
+shake128 : {d, m} (fin m) => [m] -> [d]
```
marsella, line 34: note: previously, this API defaulted to returning `[inf]`, so many of the calls to `shake128` and `shake256` are like ```take`{123} (shake128 ...)``` (others do some operations over the infinite stream first). Those `take`s can be removed now, since we can specify the length parameter directly on `shake128`. I didn't do that in this PR, but can either do it here or make a separate issue if reviewers think it'd be nice to make the calls more consistent / simple.

marsella:
> It kind of feels like something is wrong the with `sha3` if `join (toBytes ...` is needed every time its run. I've struggled quite a bit with SHA3's bit reorganization routine and haven't yet found an elegant way to avoid asking the user to rejigger values at the function boundary. If you consider updates here, please keep in in the loop.

Hm, that's a good point. I'll make a follow-up issue about it. It actually seems like we could package up the `to/fromBytes` calls into the SHA functions for any non-infinite return type, so at least the caller wouldn't have to do it -- that would cover most use cases in here except for some functions that operate over infinite streams out of SHAKE. Technically SHAKE is supposed to take a finite length parameter so I could dive into those (old versions of Dilithium, mostly) to see if we could refactor them and add a `fin d` constraint to the output length of `Keccak`.

I also have on my list to look more carefully at the state representation in this implementation. As you mention, the spec has a fairly weird indexing scheme and I'm not exactly sure that we're implementing that fully faithfully. I wonder if getting that in line would solve the input/output format inconsistency.

Primitive/Keyless/Hash/keccak.cry
marsella: Added #138.

Primitive/Keyless/Hash/keccak.cry
```diff
@@ -1,119 +1,299 @@
...
+    type c : #
+    type constraint (fin c, c < b, c > 0)
+
+/**
+ * Keccak family of sponge functions.
+ *
+ * At this time, the implementation is not generic with respect to the sponge
+ * construction; this implementation "inlines" the sponge algorithm (Alg 8)
+ * with the specific functions for Keccak (Sec 5.2).
+ * [FIPS-202] Section 4, Algorithm 8; instantiated as in Section 5.2.
```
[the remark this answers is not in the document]
marsella: To clarify, if you're reviewing and you think that would be nice, please comment here so I know to make a follow up issue.

Primitive/Keyless/Hash/keccak.cry
marsella: I made #139

Primitive/Keyless/Hash/keccak.cry
```diff
@@ -1,119 +1,299 @@
...
+    // Step 1.
+    P = M # pad `{x = r, m = m}
+    // Step 2.
+    type n = (m + 2) /^ r
+    // Step 3. `c` is a parameter for this module.
+    // Step 4. Ps = P_0, ..., P_{n-1}.
+    Ps = split P : [n][r]
+    // Step 5.
+    S = zero : [b]
+    // Step 6. We create a list `Ss` instead of overwriting the variable `S`.
```
b13decker, line 75: **nit-pick (documentation):** Maybe note Step 6 is also Step 10?

Primitive/Keyless/Hash/keccak.cry
```diff
@@ -1,119 +1,299 @@
...
+    Ss = [S] # [Keccak_f (S' ^ (Pi # (zero : [c]))) | Pi <- Ps | S' <- Ss]
+    // Step 7 - 9. This step is sometimes known as "squeeze".
+    extend : [b] -> [inf]
+    extend Z = (take`{r} Z) # extend (Keccak_f Z)
```
b13decker, line 79: **nit-pick (doc question):** I'm assuming all readers will know Cryptol enough, but maybe note at `Trunc` is natively handled by `take`?

b13decker: I really like how you note the steps and decompose the Cryptol to match NIST. Makes the visual inspection much easier.

Primitive/Keyless/Hash/keccak.cry
```diff
@@ -159,6 +162,22 @@ private
...
+    Keccak_f : [b] -> [b]
+    Keccak_f S = S'
+        where
+            A0 = unflatten S
+            rounds = [A0] # [ Round RC A | RC <- RCs | A <- rounds ]
+            S' = flatten (rounds ! 0)
+
+
+    Round : State -> State -> State
+    Round RC A = ι RC (χ (π (ρ (θ A))))
```
b13decker, line 254: **nit-pick:** Might consider swapping the order of the parameters for iota since NIST has `A` come first and then `RC`.

It would make this line of code look even more similar to NIST (granted it only took me a minute to visually inspect and determine correctness).

Primitive/Keyless/Hash/keccak.cry
marsella: Ooh, thanks. Actually Step 6 is separate -- 1-6 are "absorb" where you compute a state, and then 7-10 is "squeeze" where you recursively take some bits and call `Keccak` again until it's long enough. In squeezing there's no padding or XORing, just appending. But I'll fix the docs on L77 since Step 10 is included there!

Primitive/Keyless/Hash/keccak.cry
marsella: Thank you! I realized it wasn't super obvious that Steps 7-10 are defined here and called above, so I'll pull down the call to this function and label the two different `take`s.

Primitive/Keyless/Hash/keccak.cry
marsella: I'm going to punt this to #138 because technically the `RC` parameter doesn't match the spec at all -- `RC` is like a pre-computed state array and `ir` is the round index that should be used to compute `RC`.

Primitive/Keyless/Hash/SHA3/SHA3.cry
```diff
@@ -1,12 +1,12 @@
...
 parameter
   type digest : #
   type constraint (fin digest, digest >= 224, digest <= 512)
 
-type total = 1600
-
 sha3 : {n} (fin n) => [n] -> [digest]
 sha3 M =
-  take`{digest} (Keccak `{r = total - 2 * digest, c = 2 * digest} (M # 0b01))
+  take`{digest} (Keccak `{r=1600-(2*digest)} (M # 0b01))
```
mccleeary-galois: nit: I am happy with how this ended up in the final product but it was a bit confusing to follow commit by commit here (ie is `total = b` why is this now hard coded).

Primitive/Keyless/Hash/keccak.cry
```diff
@@ -3,7 +3,7 @@
  *
  *
  *
- * [FIPS-202]: National Institute of Standards and Technology. SHA-3 Standard:
+ * [FIPS-02]: National Institute of Standards and Technology. SHA-3 Standard:
```
mccleeary-galois: nit: same as above

Primitive/Keyless/Hash/keccak.cry
```diff
@@ -16,43 +16,75 @@
...
+ */
+type w = b / 25
+
+/**
+ * TODO: Figure out what this represents for the docs.
+ * [FIPS-202] Section 3.1.
+ */
+type ell = lg2 w
+
+type State = [5][5][w]
```
mccleeary-galois: Thank you for adding types like this, makes it much more readable.

mccleeary-galois (approved):
Good step in the right direction, as the diff is getting a bit unwieldy I would prefer to get this in and punt most things to the next PR as you had stated.

Primitive/Keyless/Hash/keccak.cry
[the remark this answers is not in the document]
mccleeary-galois: Lets create an issue to track it, but agreed on punting.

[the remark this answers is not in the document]
marsella: Added #140

marsella:
> @weaversa: It kind of feels like something is wrong the with `sha3` if `join (toBytes ...` is needed every time its run. I've struggled quite a bit with SHA3's bit reorganization routine and haven't yet found an elegant way to avoid asking the user to rejigger values at the function boundary. If you consider updates here, please keep in in the loop.

I made #142 that introduces what I think is a nicer API for interacting with the SHA3 functions. Biggest caveat is that it's not quite as obvious how it correlates with the spec. Would appreciate your input on whether it aligns with your expectations / hopes for this spec!

Issue #138, opened 2024-09-10
Clean up function implementations and state for SHA3


marsella:
Right now, the low-level functions (particularly the state transformations that make up the round function) do not look like the definitions in the spec ([see comment](https://github.com/GaloisInc/cryptol-specs/pull/137#discussion_r1750671479)).

A possibly-related issue is that the output from the SHA-3 functions is not exactly in the right order - we have to call the `fromBytes` methods to get the expected strings -- this function does some funky reversing of bit order within bytes of the output.

This is all kind of weird, and it may be that bringing the state transformations into more precise line will address the strange bit-ordering we have in the output.

- [x] Update the state transformation functions (θ, ι, χ, π, ρ, RC and RCs) to look like the spec (see #141)
- [x] Review the state-to-string transformation functions (`flatten` and `unflatten`) to make sure they're aligned with the spec (see #153)
- [x] Try to identify the source of the weird output bit ordering; determine whether it can be addressed in this PR or if it should be handled elsewhere.
    - [x] If this PR, update all the calls to SHA and SHAKE functions to make sure everything still builds and works correctly. This should remove the `toBytes` and `fromBytes` calls and maybe replace the `take`s with type parameters on shake. Moved this to #152

marsella:
I learned a bunch about bit ordering in Keccak. Essentially, the Keccak functions interpret bit strings as a sequence of bytes, where the bytes are in MSB order but the bits in each byte are in LSB order. This was previously being done outside the SHA3 functions with the `toBytes` and `fromBytes` functions.

It's not explicit in the spec when you're supposed to do this transformation: there's no conversion function called anywhere in the definition of the hash functions or XOF, which take a message and call the internal Keccak sponge. There is a conversion function from hex strings to Keccak-ordered bit vectors. However, the NIST-provided [CAVP test vectors](https://csrc.nist.gov/projects/cryptographic-standards-and-guidelines/example-values#Hashing) have inputs and outputs in plain old MSB order.

I'm currently thinking about the right API for this. I think it's more practical for the SHA functions to interact externally with normally-formatted (MSB) inputs and outputs. It's fairly easy to do this by just calling the conversion functions inside the definitions of SHA and SHAKE. However, this won't be exactly precise to the spec itself: the conversion functions are defined over even-number-of-digits hex strings and require the caller to specify the "actual" length of the string. I think this would be very unwieldy for most applications because
[...]

PR #141, opened 2024-09-18
SHA3: clean up round functions

marsella:
Addresses part of #138.

This brings the round functions into closer alignment with the spec. I found that the implementation was significantly (> 8x) slower than before, so I added hardcoded versions (and equivalence properties!) for two of the functions that involved constants. This addressed most of the slowdown.

I don't think there's anything crazy happening here; I'm just looking for general [gold-standard review](https://github.com/GaloisInc/cryptol-specs/wiki/Reviewing-guidelines).


Primitive/Keyless/Hash/keccak.cry
```diff
@@ -149,18 +149,30 @@ private
     /**
      * One of the step mappings that's part of a round of Keccak-p.
      *
+     * The effect of thuis mapping is to rotate the bits of each lane by a
```
b13decker: (no action, reviewer note) : Typo is fixed in a later commit

Primitive/Keyless/Hash/keccak.cry
```diff
@@ -189,13 +189,18 @@ private
     /**
      * One of the step mappings that's part of a round of Keccak-p.
      *
+     * The effect of this mapping is to XOR each bit with a non-linear function
+     * of two other bits in its row.
      * [FIPS-202] Section 3.2.4.
+     *
+     * Note: in the first line, XOR 1 on a single bit is equivalent to bitwise
+     * NOT. Since we operate over the whole slice defined by each z-coordinate,
+     * it's nicer to use `~`. You can see the NOT gates in Figure 6.
```
b13decker, line 239: Excellent comment! Makes it so obvious, which is perfect.

b13decker (approved):
Great comments and commit decomposition. That made the review very smooth. I didn't see any issues. I agree that the functions better match the specs now, and the properties to prove equivalence with the hardcoded fast versions are excellent. No changes requested.

PR #142, opened 2024-09-18
SHA3: add new API that handles bit ordering conversions

marsella:
This addresses part of #138, but does not complete it.

Keccak uses a weird bit ordering (MSB bytes, LSB bits). The spec is not clear about when inputs are supposed to be transformed into this representation. The previous implementation assumed that inputs would be provided in that bit ordering, so every single call site had to transform from MSB bits to Keccak order. That's so user unfriendly that I decided to make a separate API that takes inputs and returns outputs in MSB order and implicitly does the transformation to and from Keccak bit ordering. This doesn't obviously match the spec but it matches (in my opinion) the expectations of the users.

The spec does have some relatively special-purpose functions for converting between these two bit orderings (`h2b` and `b2h`). I implemented those for completeness and proved equivalence to my transformation function.

At this point, there are two different APIs for the SHA3 functions: one uses Keccak order and one uses MSB order. I've modified all the tests to use the MSB order version and plan to deprecate the Keccak order one in a subsequent PR (and change all the call sites at that point).

This is a big PR and includes a lot of stuff not really in the spec because (in my opinion) the spec is not particularly clear about how to handle this bit ordering transformation. I'm looking for feedback on whether the docs now tell the story clearly enough to distinguish, and for confirmation that the API I landed on is one you would expect.

---
Outdated: Remaining tasks:
- [x] Make sure the top level docs of the bit order file explain the intended use -- added docs throughout
- [x] Fix all the call sites of shake128 to not manually flip bits (look for tests to run where possible) -- pushing this to a separate PR that will remove the old API
- [x] document provenance of all the shake128 test vectors
- [x] consult potential reviewers: should I propagate this change to all the sha3 functions here or do that in a separate PR? -- we decided a separate PR to deprecate the old API.


Primitive/Asymmetric/Cipher/Kyber/3.01/specification.tex
marsella, line 1: I'm sorry about this diff. The only non-whitespace changes are the copyright notice at the top and the rename of the utils file on line 913.

Primitive/Asymmetric/Signature/Dilithium/Round1/DilithiumInteger.cry
marsella, line 1: Again, this is mostly whitespace. Meaningful changes are the copyright notice and the rename of utils -> `KeccakBitOrdering`.

Primitive/Asymmetric/Signature/Dilithium/Round2/DilithiumParameterized.cry
marsella, line 1: Same story here re: whitespace and renames.

Primitive/Keyless/Hash/SHA3/SHA3.cry
```diff
@@ -27,6 +28,27 @@ parameter
...
+ * `keccak` module.
+ *
+ * This expects input and output in MSB order. It handles conversion to and
+ * from the bit ordering expected by `Keccak`. If input is provided in the
+ * format described in [FIPS-202] Appendix B (a hex string with an even
+ * number of digits and an actual length `n`), use the `KOB::truncate` function
+ * to get the correct input:
+ */
+hash : {n} (fin n) => [n] -> [digest]
+hash M = KBO::reverseBitOrdering (Keccak`{d=digest} ((KBO::reverseBitOrdering M) # 0b01))
```
marsella, line 54: Intention here is to make a follow-up PR that replaces all the calls to `sha3` with calls to `hash` (and similarly for the shake functions). I'll make a follow-up issue.

Primitive/Keyless/Hash/SHA3/Tests.cry
```diff
@@ -12,111 +23,147 @@
...
-  property t1 = join (toBytes (sha3 [])) ==
-    0x6b4e03423667dbb73b6e15454f0eb1abd4597f9a1b078e3f5b5a6bc7
-
-  /*
-   * ```repl
-   * :prove t2
-   * ```
-   */
-  property t2 = join (toBytes (sha3 0b11001)) ==
-    0xffbad5da96bad71789330206dc6768ecaeb1b32dca6b3301489674ab
```
marsella, line 34: It was only after I removed all of these short bitwise tests that I realized they're from [the CSAG example files](https://csrc.nist.gov/projects/cryptographic-standards-and-guidelines/example-values). Anyway I think the CAVP ones are nicer but I can re-add these back in if you'd like.

Primitive/Keyless/Hash/KeccakBitOrdering.cry
```diff
@@ -0,0 +1,313 @@
...
+ */
+toBytes : {n} (fin n) => [8*n] -> [n][8]
+toBytes s = reverse (split (reverse s))
+
+/**
+ * Deprecated!!
+ * Joins a list of bytes into a list of bits, using little-endian bit order.
+ */
+fromBytes : {n} (fin n) => [n][8] -> [8*n]
+fromBytes bs = reverse (join (reverse bs))
```
marsella, line 26: note: this part came from `Primitive/Keyless/Hash/utils.cry`. It's still used by lots of files so we're keeping it for now.

marsella:
I started playing around with using this API in some of the dependent specs. The main drawback so far is that those specs largely work over byte arrays, like of type `[m][8]`, and so we have to do something at the call site like ```groupBy`{8} (shake128 (join input))```

I'm going to try to poke around and see if I can come up with a type signature that would allow bit or byte arrays.

marsella:
> I'm going to try to poke around and see if I can come up with a type signature that would allow bit or byte arrays.

It looks like there's not a way to support both options with one function. I think the two reasonable options are:
1. Offer one bit-wise function and make the caller do the `groupBy` and `join`.
2. Offer a bit-wise function `xof : [m] -> [d]` and a byte-wise function `xofBytes : [m][8] -> [d][8]`.

It might be best to leave as-is for now and wait until the PR where I replace all the uses of the old API with this new API. Then we can see whether bits or bytes is more common.

Primitive/Keyless/Hash/KeccakBitOrdering.cry
```diff
@@ -0,0 +1,313 @@
...
+ * and back.
+ *
+ * We pad the strings using our `reverseBitOrdering` function; the padding goes
+ * to the "inside" of the last byte. This matches the convention in the CAVP
+ * test vectors. E.g. if `-` is the actual message and `P` is the padding, a
+ * vector with 3 bits of padding looks like:
+ * `-------- -------- -------- PPP-----`
+ *
+ * The type parameters are:
+ * - `m`: the number of pairs of hex digits in the input string
```
mccleeary-galois, line 276: Does it make sense to test any thing other than 3 pairs in the future?

Primitive/Keyless/Hash/SHA3/SHA3.cry
mccleeary-galois: Let's make sure to link the issue here.

mccleeary-galois (approved):
LGTM, I agree with the choice you made to make the Keccak transformation implicit from the top level.

Primitive/Keyless/Hash/KeccakBitOrdering.cry
marsella: Discussed offline -- I'll add a note explaining why 3 (the reason is that there's no reason, I just picked any number to show off the more interesting parameter of the padding length) and noting that the tests should pass with any pair-count.

Primitive/Keyless/Hash/SHA3/SHA3.cry
marsella: I added #152.

marsella:
I made the change that @mccleeary-galois requested and also separated a commit that renamed + changed a file into two (a6bba748a73072cedd59df56f6ec217dc9255e4e and https://github.com/GaloisInc/cryptol-specs/pull/142/commits/f3833bdcf2fe0acfcee04fb1e838b0d028588657) in case @sauclovian-g is lurking.

PR #150, opened 2024-10-03
Add ML-KEM KATs

### LINKED ISSUE #148: Add easy-to-run KATs to ML-KEM
There are a bunch of small modifications we need to make to ML-KEM to bring it up to our gold standard (see e.g. #135 #144 #145 #146 #147). ML-KEM currently can be run against the KATs (see `kat.awk`) if you download one of the `rsp` files and run it against the implementation. However, for faster iteration it would be nice to have a smaller `Tests` section with a few of the KATs hard-coded, so we can quickly load and `:check-docstrings`.

There are not official KATs for the final draft yet, but there are [some candidate KATs](https://github.com/post-quantum-cryptography/KAT/tree/main/MLKEM) available to test with.

- [x] Reorganize ML-KEM directory to match the other directories we have (add `Instantiations/` directory for `ml_kem*.cry` and a separate `Tests/` directory -- see `Common/EC/PrimeField/` for an example)
- [x] Add a test file for each of the three instantiations. Put 2-3 KATs in each one. Make sure to use the final draft version. Add docs to cite the sources of the KATs.

marsella:
Closes #148 

This adds a few easy-to-run tests for ML-KEM that will hopefully make it easier to be confident about all the upcoming changes in the ML-KEM gold standard issues. They are not fast and I only added one for each parameter set, although there are many more in the sourced repo.

In addition to the work listed in the linked issue, this also renames all the modules around here to adhere better to our standards.


mccleeary-galois (approved):

sauclovian-g:
I would strongly recommend doing just the ```git rename``` ops as one commit, and putting all the content changes in a second separate one, even though that probably means the first commit won't be buildable. Otherwise you're at the mercy of git's rename detection heuristics. (Note that if you look at the first commit here, it mostly comes up as deletes and entirely new files; that's because the rename detection heuristics don't work very well. 😐 This means for example that you won't be able to cross the rename with git log or git blame.)

At this point that'll be kind of a pain to retcon in though and maybe it's not really worthwhile...

marsella:
> I would strongly recommend doing just the `git rename` ops as one commit, and putting all the content changes in a second separate one, even though that probably means the first commit won't be buildable. Otherwise you're at the mercy of git's rename detection heuristics. (Note that if you look at the first commit here, it mostly comes up as deletes and entirely new files; that's because the rename detection heuristics don't work very well. 😐 This means for example that you won't be able to cross the rename with git log or git blame.)
> 
> At this point that'll be kind of a pain to retcon in though and maybe it's not really worthwhile...

It didn't take that long to fix, so now the individual commit diffs look good. The overall diff still fails to recognize the rename, but hopefully this will help anyone pawing through the history.

PR #153, opened 2024-10-09
simplify state-to-bit-string conversions in SHA3

### LINKED ISSUE #138: Clean up function implementations and state for SHA3

marsella:
Closes #138!

This attempts to bring the conversions between the state matrix and the state bit string from [the SHA3 spec](https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.202.pdf) closer into alignment with the spec. 

I'm not entirely satisfied with it because there's a `reverse` in there; I don't understand why it's necessary. Basically operating on "slices" / the z-indexed vectors happens in the opposite order than I expected it to; I haven't found a clear explanation as to why. I added a half-baked explanation in the docs but I'm not very satisfied with it.

The alternative to the presentation here would be to add explicit z-indexing in both the `flatten` function and in all the step mappings. This would remove the spurious `reverse` and generally would bring everything into closer alignment with the spec. It might affect speed; I'm not sure. It might not work, but I suspect it would. It would (subjectively) make the step mappings a bit uglier (but then, lovely code is not exactly the point of these implementations).

Request for reviewers: do you think I should try to switch to no-reverse and explicit z-indexing?

In testing this, I found several doc strings that were not correctly commented as such; I fixed that throughout the repo.

### DIFF
--- a/Primitive/Keyless/Hash/keccak.cry
+++ b/Primitive/Keyless/Hash/keccak.cry
@@ -131,6 +131,20 @@ private
         Lanes = map reverse (split`{25} S)
 
     /**
+     * Convert a string into a state array, but without indexing.
+     * Equivalent to [FIPS-202] Section 3.1.2.
+     */
+    unflatten_noindex : [b] -> State
+    unflatten_noindex S = transpose (groupBy (reverse (groupBy`{w} (reverse S))))
+
+    /**
+     * ```repl
+     * :prove unflattens_match
+     * ```
+     */
+    property unflattens_match S = unflatten S == unflatten_noindex S
+
+    /**
      * Convert a state array into a string.
      * [FIPS-202] Section 3.1.3.
      */
@@ -145,6 +159,20 @@ private
         Planes = [ Lanes@0@j # Lanes@1@j # Lanes@2@j # Lanes@3@j # Lanes@4@j
             | j <- [0..4]]
         S = join Planes
+
+    /**
+     * Convert a state array into a string, but without indexing.
+     * Equivalent to [FIPS-202] Section 3.1.3.
+     */
+    flatten_noindex : State -> [b]
+    flatten_noindex A = reverse (join (reverse (join (transpose A))))
+
+    /**
+     * ```repl
+     * :prove flattens_match
+     * ```
+     */
+    property flattens_match A = flatten A == flatten_noindex A
 
     /**
      * One of the step mappings that's part of a round of Keccak-p.
@@ -378,11 +406,11 @@ private
             // The round transformation is defined in [FIPS-202] Section 3.3.
             Rnd A ir = ι (χ (π (ρ_hardcoded (θ A)))) ir
             // Step 1.
-            A0 = unflatten S
+            A0 = unflatten_noindex S
             // Step 2.
             rounds = [A0] # [ Rnd A ir | A <- rounds | ir <- [0..nr - 1]]
             // Step 3.
-            S' = flatten (rounds ! 0)
+            S' = flatten_noindex (rounds ! 0)
 
 /**
  * See https://keccak.team/files/Keccak-reference-3.0.pdf, Section 1.2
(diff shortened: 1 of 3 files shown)


Primitive/Keyless/Hash/keccak.cry
```diff
@@ -115,16 +115,36 @@ private
...
+     * `(x, y, z)`. Since the lanes determined by a pair `(x, y)` are composed
+     * of consecutive bits, we don't index into them separately; instead, we
+     * separate `S` into 25 lanes of length `w` and then place those lanes in
+     * the correct order according to the `(x, y)` coordinates.
+     * For ease of implementation of the subsequent step mappings, the bits of
+     * each lane are reversed.
      */
     unflatten : [b] -> State
-    unflatten p = transpose (groupBy`{5} (reverse (groupBy`{w} (reverse p))))
+    unflatten S = [[ Lanes@((5 * y + x))
```
marsella, line 128: note: these names (flatten, unflatten) don't actually appear in the spec. I am open to suggestions of alternative names here, if there is anything that makes more sense to you.

Primitive/Keyless/Hash/keccak.cry
```diff
@@ -115,16 +115,36 @@ private
...
      * Convert a string into a state array.
      * [FIPS-202] Section 3.1.2.
+     *
+     * Note: The spec describes this in terms of all three coordinates
+     * `(x, y, z)`. Since the lanes determined by a pair `(x, y)` are composed
+     * of consecutive bits, we don't index into them separately; instead, we
+     * separate `S` into 25 lanes of length `w` and then place those lanes in
+     * the correct order according to the `(x, y)` coordinates.
+     * For ease of implementation of the subsequent step mappings, the bits of
+     * each lane are reversed.
```
marsella, line 125: note: I feel like there must be a cryptol-implementation-related explanation for this but I don't know what it is.

Primitive/Keyless/Hash/keccak.cry
mccleeary-galois, line 128: We want to try and avoid indexing if we can. This is also probably why it was implemented this way.

marsella: Gotcha. I like this because it's more obvious relative to the spec, but I can add the other one back in and prove equality so we're not indexing in practice.

mccleeary-galois: I am good with that approach, as I agree that the implementation here was much easier to follow along with in the spec.

Primitive/Keyless/Hash/keccak.cry
mccleeary-galois: I am wondering if the reverse is so that the indexing wasn't necessary? I need to think that through a bit more.

Primitive/Keyless/Hash/keccak.cry
marsella: Fixed this!

mccleeary-galois (approved):

PR #154, opened 2024-10-10
Separate + document K-PKE component scheme

### LINKED ISSUE #145: Separate K-PKE scheme in ML-KEM
Right now the ML-KEM scheme has all of its functions inline. The K-PKE scheme is a sub-scheme that's used to implement ML-KEM, but cannot be used on its own. It would be good to separate these functions into their own submodule so that they are more obviously self contained, not the ML-KEM scheme, and not available for use outside of the module.

While we're moving around all the K-PKE stuff, we should also make a point to bring it up to [our gold standard](https://github.com/GaloisInc/cryptol-specs/wiki/Reviewing-guidelines). Some suggestions are here, but this list isn't necessarily exhaustive.

- [x] Make a K-PKE submodule in the ML-KEM specification. Remove the `K_PKE_` prefix on the functions and update the calls to use the module path (`K_PKE::KeyGen`) instead.
- [x] Write top-level docs for submodule explaining the limitations (doesn't do any input validation) and with loud warnings against using this outside the ML-KEM context.
	- [x] Also make sure function-level docs have all appropriate warnings.
- [x] Update the code to more obviously match the spec (there's not that much to do, I think)
	- [x] Move the XOF call into `SampleNTT`
	- [x] Consider making an `EncryptionKey` and `DecryptionKey` type to use in type sigs
	- [x] Add inline docs to label the steps of each algorithm

marsella:
Closes #145 

This pulls the K-PKE scheme from [FIPS 203, Section 5](https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.203.pdf) into its own module, so it's harder to misuse and easier to see the containment. It's really just a code clarity PR, I didn't find any new bugs. Added a few named types, removed some tuples to match the style guide in my head, and added lots of whitespace and inline docs.

Reviewer requests: Please make sure this matches our [gold standard criteria](https://github.com/GaloisInc/cryptol-specs/wiki/Reviewing-guidelines). In particular, if there are any bits that seem like inconsistent styling or if any names don't quite match up with the spec, please flag them.


Primitive/Asymmetric/Cipher/ML_KEM/Specification.cry
```diff
@@ -723,88 +724,149 @@ dotMatMat matrix1 matrix2 = transpose [dotMatVec matrix1 vector | vector <- m']
...
-  m' = K_PKE_Decrypt(sk, c)
-
+ * ⚠️ This scheme is not approved for stand-alone use! ⚠️
+ * K-PKE is an encryption scheme consisting of three algorithms `(KeyGen,
+ * Encrypt, Decrypt)`, which are used to instantiate the approved ML-KEM
+ * scheme. It's not secure as a standalone scheme; it doesn't do any input
+ * checking.
+ * [FIPS-203] Section 5.
+ */
+submodule K_PKE where
```
marsella, line 736: todo: check whether submodules are private by default. If not, make this explicitly private.

Primitive/Asymmetric/Cipher/ML_KEM/Specification.cry
```diff
@@ -325,9 +325,10 @@ property CorrectnessEncodeDecode' f = Decode'`{12}(Encode'`{12} f) == f
  *
  * [FIPS-203] Section 4.2.2, Algorithm 7.
  */
-SampleNTT : [inf]Byte -> Z_q_256
-SampleNTT b = take elements
-    where elements = SampleNTTInf b
+SampleNTT : [34]Byte -> Z_q_256
+SampleNTT B = take elements where
+    C_stream = XOF(B)
+    elements = SampleNTTInf C_stream
```
nlschimanski, line 331: In this implementation, SampleNTT is broken into two functions: SampleNTT and SampleNTTInf. The specification defines SampleNTT as a single algorithm, so it took me a bit to match them up. Is this preferred? If so, maybe an explanation as to why they are broken into two functions would be helpful.

Primitive/Asymmetric/Cipher/ML_KEM/Specification.cry
```diff
@@ -723,88 +724,149 @@ dotMatMat matrix1 matrix2 = transpose [dotMatVec matrix1 vector | vector <- m']
...
+        // Step 16.
+        s_hat = NTT s
+        // Step 17.
+        e_hat = NTT e
+        // Step 18.
+        t_hat = (dotMatVec A_hat s_hat) + e_hat
+        // Step 19.
+        ekPKE = (Encode`{12} t_hat) # ρ
+        // Step 20.
+        dkPKE = Encode`{12} (s_hat)
```
nlschimanski, line 783: Having the steps of the algorithm annotating the code is super helpful!!

Primitive/Asymmetric/Cipher/ML_KEM/Specification.cry
```diff
@@ -723,88 +724,149 @@ dotMatMat matrix1 matrix2 = transpose [dotMatVec matrix1 vector | vector <- m']
...
+        // Step 17. In the spec, the second parameter is `N = 2k`. In this
+        // implementation, `N` itself is out of scope, so we use the fixed
+        // value instead.
+        e2 = SamplePolyCBD`{eta_2} (PRF r (2 * `k))
+        // Step 18.
+        y_hat = NTT y
+        // Step 19.
+        u = NTTInv (dotMatVec (transpose A_hat) y_hat) + e1
+        // Step 20.
+        mu = Decompress'`{1} (DecodeBytes'`{1} m)
```
nlschimanski, line 819: Should 'DecodeBytes' be named 'ByteDecode' instead so it matches the naming convention used in the spec? Similar for 'EncodeBytes' below.

Primitive/Asymmetric/Cipher/ML_KEM/Specification.cry
```diff
@@ -723,88 +724,149 @@ dotMatMat matrix1 matrix2 = transpose [dotMatVec matrix1 vector | vector <- m']
...
+     * ```repl
+     * :set tests=3
+     * :check CorrectnessPKE
+     * ```
+     */
+    CorrectnessPKE : [32]Byte -> [32]Byte -> [32]Byte -> Bit
+    property CorrectnessPKE d m r = (m' == m) where
+        (pk, sk) = KeyGen d
+        c = Encrypt pk m r
+        m' = Decrypt sk c
```
nlschimanski, line 869: I saw that this correctness property is stated in the spec. Would it be useful to add a reference here to where it shows up in the spec? (Or maybe add a reference pointing to the paper where it's proved?)

nlschimanski (commented):
I didn't notice a spec written in Literate Cryptol in the ML_KEM folder but I do see some documentation describing the protocol and a copy of the spec. Is it important to have the full specification written in Literate Cryptol in LaTeX for this implementation or is the current set of documentation sufficient?

I focused on the changes made to 'Specification.cry' in my review against the Gold standard criteria rather than the broader set of ML_KEM files. So, let me know if a deeper review is needed for this PR.

marsella (commented):
> I didn't notice a spec written in Literate Cryptol...

You reviewed the right thing, thank you! At this time we are focusing on getting working specs for many standardized algorithms first, and we will make them literate later.

Primitive/Asymmetric/Cipher/ML_KEM/Specification.cry
nlschimanski: It's mentioned on page 14 of section 3.2 and I think the proof of the correctness is shown in reference 14.

marsella: Hm, all I see there is the properties of the ML-KEM parent scheme, not this K-PKE component scheme. I'll update the docs here a bit but I don't think there's a specific writeup of the properties of an encryption scheme anywhere in this spec.

nlschimanski: Ah, yes! You're right, it for ML-KEM rather than the K-PKE component.

nlschimanski (approved):

Issue #144, opened 2024-10-01
Bring encoding and compression functions in ML-KEM up to gold standard


marsella:
There are some differences between our implementation and [the FIPS 203 spec](https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.203.pdf) that should be addressed to make it very obvious that our spec is correct. There are several missing properties and the encode/decode functions need to be a bit more explicit to map correctly onto the spec.

The current state of encoding / decoding is a little confusing. The spec defines `ByteEncode` and `ByteDecode` , which basically convert between `d`-bit integers and byte arrays, for $d \in [1, 12]$. The behavior for `d=12` is slightly different than for other values; instead of 12-bit integers, the conversion is to integers mod `q`.  We can implement this using numeric constraint guards (we do actually need multiple implementations). Right now there are several different encode / decode functions that I think are used to separate these cases (see `DecodeBytes` that returns an `ell`-bit integer and `Decode` that returns an array of integers mod `q`)

Also, the spec overloads the function names (both for encode/decode and compression) and applies them to both single inputs and arrays of inputs; we can't overload names in Cryptol so we need an alternate solution. Since this is an issue across multiple functions (including these and NTT at least), I propose we add suffixes to the applications on arrays `_Vec` and on matrices `_Mat`. So we might have the following functions:
```
Compress : {d} (fin d) => Z q -> [d]
Compress_Vec : {d, n} (fin d) => [n](Z q) -> [n][d]
Compress_Mat : {d, n, m} (fin d) => [n][m](Z q) -> [n][m][d]
```

These are the specific tasks I identified while auditing / making #143. If other details arise while making these changes, please note them in the comments.

- [x] Add the bit/byte conversion example on P21 as a property
- [x] Add the missing compression correctness property from P21. We have one, but we still need to add "Compress_𝑑(Decompress_𝑑(𝑦)) = 𝑦 for all 𝑦 ∈ ℤ_2𝑑 and all 𝑑 < 12."
- [x] Write a single `ByteEncode` that uses numeric constraint guards to handle all cases of `d` (the output type will be something like `Z (min (2^^d) q)`)
- [x] Add properties for encoding and decoding
	- [x] For `d < 12` they are each others' inverse on all inputs, and in each direction (e.g. `encode (decode x) == x` and `decode (encode x) == x`. 
[...]

PR #155, opened 2024-10-11
MLKEM: Bring sampling + crypto functions up to gold standard

### LINKED ISSUE #146: Bring sampling and cryptographic functions in ML-KEM up to gold standard
This is a bunch of small clean-up tasks that arose during implementation of #143. They're all related to pseudo-randomness and sampling. See [ML-KEM spec](https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.203.pdf), [gold standards](https://github.com/GaloisInc/cryptol-specs/wiki/Reviewing-guidelines)

I also think that there is some un-aligned bit ordering in here that's caused by SHA3. The SHA-3 spec expects bits in an unusual order (MSB bytes, LSB bits) for both input and output; previously, the SHA-3 interface required the user to pass / take their input / output in that order. As of #142, there is a sleeker SHA-3 interface that interacts with MSB bit vectors. However, ML-KEM assumes bit vectors are in LSB order. There is some strange usage of SHA3 in ML-KEM: the uses of SHAKe128 and SHAKE256 convert from MSB to SHA3 bit vectors, but returns SHA3 bit vectors. ~I'm not sure whether there are additional bit order changes in some of the other bit-manipulation functions, like `BytesToBits`, `BitsToBytes`, and maybe `the BitstoZ` and `ZtoBits`~ (see [comment for notes on these conversions](https://github.com/GaloisInc/cryptol-specs/issues/144#issuecomment-2400678743)). The gold-standard version of this spec should use the gold-standard version of SHA-3, so we will need to track down all the places where bit ordering is compensating for SHA-3 and remove them from this spec.

- [x] Replace the parameterized import of SHA3 with concrete imports of `SHA3-256` and `SHA3-128`
- [x] Update uses of SHA3 (hash and xofs) to use the new API that doesn't require reversing bits (related: #138, #142)
  - [x] Fix any other bit ordering functions that propagate this issue.
- [x] PRF: Tighten the constraint on the `eta` parameter (should only be 2 or 3). Make sure it aligns with the spec (byte count, not bit count). Maybe just call it `eta`. 
- [x] SampleNTT: make `inf` version a function in the `where` instead of a standalone thing. Try to simplify the control flow to look more like the spec. Explain where laziness trumps spec-matching (e.g. there is no `j`.)
- [x] SamplePolyCBD: Tighten bound on `eta`. Update naming & details to look more like the spec. Adjust spacing to be more consistent with Marcella's preferred standards 🙃 (someday perhaps soon I'll write a style guide).

marsella:
Closes #146 

This cleans up and documents the functions related to sampling various types of polynomials, and the PRFs and hash functions that support said sampling in ML-KEM. 

That issue has some discussion about bit ordering. Luckily it turned out to not be as complicated as I feared for this PR; I think the byte-vector-endianness issue discussed in #144 is still around, but that can be handled independently of this. For the purposes of this PR, I was able to use the new SHA3 API without having to do any further byte rearrangement.

I don't have any specific review requests.


nlschimanski (approved):
It all looks great to me! I really like the function import capitalization convention and the SampleNTT implementation is WAY easier to follow in this form. I also noticed clear explanations in the code when it differed slightly from the spec (e.g., since Cryptol isn't imperative, recursion was used). There is also useful documentation explaining why some steps were taken when they weren't in the spec (e.g., conversions that are implicit).

Issue #147, opened 2024-10-03
Bring NTT sections of ML-KEM up to gold standard


marsella:
Right now there is a bunch of stuff related to NTT (Section 4.3) in the main body of ML-KEM. There are naive and fast versions of the NTT algorithm (with a "dispatcher" to hard-code which one to use), as well as 1D and 2D applications. We should try to contain this a little more intentionally.

There are also some places where the NTT-related code doesn't obviously align with [the spec](https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.203.pdf). 

- [x] Make an NTT submodule that contains all the NTT stuff -- everything in section 4.3. Maybe also section 4.3.1 -- I'm not sure what will make the most sense.
- [x] Add types for Rq and Tq instead of using the same Z_q_256 for both. I'm not sure if this should belong in the NTT submodule or the enclosing one.
- [x] At the "dispatcher", add docs explaining the equivalence of the two versions and note the specific place in the spec that says it's fine to use equivalent algorithms.
- [x] We need to be able to run `NTT` over vectors and `NTTInv` over single elements and vectors. Rename `NTT'` and `NTT` to `NTT` and `NTT_Vec`, and similarly for `NTTInv` (related: #144 discusses the naming choice)
- [x] In the submodule, make all the different instantiations of NTT private. The only public things in the submodule should be `NTT`, `NTT_Vec`, `NTTInv`, `NTTInv_Vec`, and, if they're included, `MultiplyNTTs` and the `R_q`/`T_q` types.
- [x] ~Update the naive version (`ParametericNTT`) to more obviously match the spec.~ Punting this to a separate issue #163
	- [x] In the spec, `BitRev7` is supposed to operate over 7-bit vectors but it currently operates over 8-bit vectors. Try to align this better with the spec.
	- [x] ~The use of `BitRev7` in `ParametricNTT` is different from what's in the spec. Fix it or add a comment explaining why they're equivalent.~ Also punting
- [x] Make sure all the NTT equivalence properties have docstrings
- [x] Align `MultiplyNTTs` better with the spec. E.g. the exponent for zeta is not the same as what's in the spec. Fix it or add a comment explaining why they're equivalent.

PR #156, opened 2024-10-16
ML-KEM: Try to clean up the NTT implementations

marsella:
Does not quite finish #147.

This addresses most of the clean-up issues in the linked issue, except for cleaning up the naive NTT version described [in the spec](https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.203.pdf) (see Algorithms 9 and 10). That has a dense nested loop structure whose bounds depend on each other's variables and that iteratively updates the state of the `f` variable. We have a version that works but doesn't clearly map onto the spec, and I spent a day failing to find a more obviously-matched way to implement it.
Update: I made a recursive version of Algorithm 9 which may or may not be better. I solicit opinions from reviewers on whether they like it or not.

Everything else should be addressed in a reasonable way. Looking for [general gold-standard quality review](https://github.com/GaloisInc/cryptol-specs/wiki/Reviewing-guidelines).

### DIFF
--- a/Primitive/Asymmetric/Cipher/ML_KEM/Specification.cry
+++ b/Primitive/Asymmetric/Cipher/ML_KEM/Specification.cry
@@ -699,11 +699,11 @@ submodule NTT where
     // Here, we can choose to call either the naive or fast NTT
     //////////////////////////////////////////////////////////////
 
-    NTT' : Rq -> Tq
-    NTT' = fast_ntt
+    NTT : Rq -> Tq
+    NTT = fast_ntt
 
-    NTTInv' : Tq -> Rq
-    NTTInv' = fast_invntt
+    NTTInv : Tq -> Rq
+    NTTInv = fast_invntt
 
     /**
      * The notation `NTT` is overloaded to mean both a single application of `NTT`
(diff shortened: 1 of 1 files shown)


marsella:
I revisited this today and have not gained any insight in the interim week. The innermost loop of NTT accesses and updates `2 * len` elements of the loop, where `len` is variable based on previous loop iterations. I was hoping I could try to fold across the inner loop, but ran into several problems:
- if the iterator `j` is a variable, I don't think there's a way to update a single element in the output list (everything I can imagine uses `take` and `drop` to preserve the unchanged elements, which would require `j` to be a type)
- There's no way to "build up" the output list as we go because the `fold` function has to be the same for every iteration, which means it can't be parameterized by an increasing length parameter.

I keep ending up in the following train of thought:
- If I want to update elements of the list one-by-one, the index of them has to be a type
- If the index is a type I'm going to have to use recursion to update it as we iterate

There might be a somewhat cleaner way to implement the recursive loops than what I did in e742abb and 0b355c3, but I think that's going to have to be the basic structure, which means it's going to be hard to read. Unless we can find a different NIST-standardized NTT description that's stated to be equivalent to this one, and can implement that instead.

Anyway, for now I propose we take the clean-up commits through b56844e and I'll spin out a separate issue to implement NTT itself more closely to the spec.

marsella:
I reverted my attempt at a recursive version and made #163 to handle fixing the NTT implementation.

mccleeary-galois (approved):
Looks like a good step in the right direction to me!

Issue #120, opened 2024-08-21
Fix README links


marsella:
Several interesting algorithms have been updated and subsequently given new names. Some of the links in the matrix for the README are now broken (SHA2 specifically, but likely others). We should fix them.

mccleeary-galois:
Still TODO after #151 

- [ ] TripleDES
- [ ] SHA256
- [ ] SHA2
- [ ] SHA3

PR #159, opened 2024-10-18
Fix broken README links.

nlschimanski:
Links to Cryptol implementations of TripleDES, SHA256, SHA2, SHA3 are fixed.

This _probably_ closes issue #120.


marsella:
Thank you! I clicked through the links in `master` and it looks like `ML-KEM` is now broken as well (I switched from a `-` to an `_`). Could you either change that here or add it to the linked issue as a future todo?

PR #160, opened 2024-10-21
ML-KEM: Bring encoding functions in line with the spec

marsella:
Addresses part of #144, but does not complete it.

This handles all the encode / decode parts of #144. It was getting a little hefty as-is so I'm breaking it into its own PR.

This was mainly a project in mucking around with types. [The spec assumes](https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.203.pdf) that all arithmetic operations can be done on all integral types, but that's not true in Cryptol. For example, you can't multiply a 1-bit vector by 2, and you can't mod an integer mod `q` by 2. These things "just happen" in the spec, but in this implementation we have to manually convert types between bit vectors and integers mod, and sometimes we have to use more width-independent operations (like bit shifting instead of multiplying by `2^j`). 

The original plan for this PR was to combine the `ByteEncode` and `ByteEncode12` functions into one (and similarly for decoding). However in order to do so I would've had to represent `d`-bit vectors as integers mod `2^d` (e.g. using the `Z` type), and that was ugly, hard to understand, and slow to prove any properties about. So I kept them separate, as they were previously, and just changed the names to be more in line with the spec.

The other overloaded name in the spec is that the encode/decode functions use the same name to apply to either one or many inputs. I added a `_Vec` suffix to distinguish these cases.

This feels very cluttered to me with so many properties and conditional functions to handle just 2 algorithms in the spec. I welcome suggestions to reduce that mess.


Primitive/Asymmetric/Cipher/ML_KEM/Specification.cry
```diff
@@ -203,129 +203,203 @@ Compress x = map Compress'`{d} x
...
+ * Decode a byte array into an array of integers mod `q`.
+ * [FIPS-203] Section 4.2.1 Algorithm 6.
+ *
+ * Note: In this implementation, we treat the `d < 12` case separately from
+ * `d == 12` because it allows us to better express the different integer
+ * types. For `d = 12`, we use the integers-mod (`Z`) type.
+ * See `ByteDecode` for the `d < 12` case.
+ */
+ByteDecode12 : [32 * 12]Byte -> [256](Z q)
+ByteDecode12 B = F' where
```
marsella, line 337: note: this function is never actually used alone, but I don't feel like it's reasonable skip it and only make `ByteDecode12_Vec`.

Primitive/Asymmetric/Cipher/ML_KEM/Specification.cry
```diff
@@ -203,129 +203,203 @@ Compress x = map Compress'`{d} x
...
- */
-regroup B = reverse (groupBy (join (reverse B)))
-
 /**
- * This is used in some places where the `ByteEncode` function is required in
- * the spec. It looks like a 2D version of it?
- */
-EncodeBytes' : {ell, c} (fin ell, ell > 0, fin c) => [c * 8][ell] -> [c * ell]Byte
-EncodeBytes' = regroup
+ * Encode an array of `d`-bit integers into a byte array, for `d < 12`.
```
marsella, line 207: todo: I was also thinking of making all these private, since they're not part of the public API of ML-KEM.

Primitive/Asymmetric/Cipher/ML_KEM/Specification.cry
```diff
@@ -203,129 +194,203 @@ Compress x = map Compress'`{d} x
...
+ * Decode `k` joined byte arrays into `k` arrays of integers mod `q`.
  */
-Decode' : {ell} (fin ell, ell > 0) => [32 * ell]Byte -> Z_q_256
-Decode' B = map BitstoZ`{ell} (split (BytesToBits B))
+ByteDecode12_Vec : [k * 32 * 12]Byte -> [k][256](Z q)
+ByteDecode12_Vec B = map ByteDecode12 (split B)
 
 /**
- * Proof that the efficient decode function is the same as the spec version.
+ * Multiplying a value by `2^^j` is the same as bit-shifting it left by `j`
```
nlschimanski, line 350: Should 2^^j be 2^j in this line?

Primitive/Asymmetric/Cipher/ML_KEM/Specification.cry
```diff
@@ -203,129 +194,203 @@ Compress x = map Compress'`{d} x
...
- * the version in the spec.
- *
- * This should be equivalent to [FIPS-203] Section 4.2.1, Algorithm 5.
+ * When `d < 12`, the byte encoding and decoding functions should be
+ * one-to-one inverses of each other.
+ * [FIPS-203] Section 4.2.1 "Encoding and decoding", first point.
+ * ```repl
+ * :prove ByteEncodeInvertsByteDecode`{1}
+ * :prove ByteEncodeInvertsByteDecode`{d_u}
+ * :prove ByteEncodeInvertsByteDecode`{d_v}
```
nlschimanski, line 372: Where does d_u and d_v come from?

nlschimanski (commented):
The EncodeBytes and DecodeBytes implementations look quite a bit different from the spec, but the explanations for why that's the case help build confidence. I especially appreciate proving each of the claims that were made (e.g., multiplying by 2^j is the same as left shift).

Primitive/Asymmetric/Cipher/ML_KEM/Specification.cry
```diff
@@ -203,129 +194,203 @@ Compress x = map Compress'`{d} x
...
+
+    // The following is the same as in `ByteEncode`. See notes above.
+    as = [[ a >> j | j <- [0..d-1]] | a <- F' ]
+    b = join [ [ a ! 0 | a <- ajs] | ajs <- as]
+    B = BitsToBytes b
 
 /**
- * This is used in some places where the `ByteEncode` function is required in
- * the spec. It's a 3D version of `EncodeBytes'`.
+ * Encode a set of `k` vectors of integers mod `q` into a byte array.
```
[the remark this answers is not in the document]
marsella: Good catch, thanks.

Primitive/Asymmetric/Cipher/ML_KEM/Specification.cry
marsella: Added an explanation here and everywhere I use these without comment in the doctests.

nlschimanski (approved):

mccleeary-galois (approved):
LGTM as well, note that #166 applies here as well.

PR #161, opened 2024-10-23
ML-KEM: Improve compression and byte conversion functions

### LINKED ISSUE #144: Bring encoding and compression functions in ML-KEM up to gold standard

marsella:
Closes #144 (assuming that #160 has already been merged).

This has two parts:
- Cleans up the compression functions. These were correct and nicely written, I just changed names and docs.
- Attempts to clean up bits-to-bytes conversion functions.

For the second part, I am honestly kind of confused. I thought the endianness was wrong but I tried rewriting it multiple times to make it look more like the spec and it seems correct. The KATs pass and all the internal properties pass. If anyone discerns something weird I'm doing with bit ordering I would love feedback.

### DIFF
--- a/Primitive/Asymmetric/Cipher/ML_KEM/Specification.cry
+++ b/Primitive/Asymmetric/Cipher/ML_KEM/Specification.cry
@@ -138,18 +138,30 @@ property roundingWorks y = y >= 0 ==> roundUpWorks && roundDownWorks where
     roundDownWorks = roundAway (y' + 0.4) == y
 
 /**
- * Compression from an integer mod `q` to an integer mod `2^d`.
+ * Compress an integer mod `q` into an integer mod `2^d`.
  * [FIPS-203] Section 4.2.1, Equation 4.7.
  */
-Compress'' : {d} (d < lg2 q) => Z q -> [d]
-Compress'' x = fromInteger(roundAway(((2^^`d)/.`q) * fromInteger(fromZ(x))) % 2^^`d)
+Compress'' : {d} (d < 12) => Z q -> [d]
+Compress'' x = y where
+    // Convert from an integer mod `q` to a rational number.
+    x' = fromInteger (fromZ x) : Rational
+    // Compress. Note that `/.` denotes division of rationals.
+    y' = roundAway (((2^^`d) /. `q) * x')
+    // mod 2^^d (by converting from an integer to a d-bit vector).
+    y = (fromInteger y') : [d]
 
 /**
- * Decompression from an integer mod `2^d` to an integer mod `q`.
+ * Decompress an integer mod `2^d` into an integer mod `q`.
  * [FIPS-203] Section 4.2.1, Equation 4.8.
  */
 Decompress'' : {d} (d < lg2 q) => [d] -> Z q
-Decompress'' x = fromInteger(roundAway(((`q)/.(2^^`d))*fromInteger(toInteger(x))))
+Decompress'' y = x where
+    // Convert from a d-length bit vector to a rational number.
+    y' = fromInteger (toInteger y) : Rational
+    // Decompress! As before, `/.` is division of rationals.
+    x' = roundAway((`q /. (2^^`d)) * y')
+    // Convert from an integer to an integer mod `q`.
+    x = (fromInteger x') : Z q
 
 /**
  * When `d` is large, compression followed by decompression must not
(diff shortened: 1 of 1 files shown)


mccleeary-galois (commented):
Caught this on this PR, but the ML-KEM proofs currently require `z3 v4.13+`. This is what is installed by default with `brew install cryptol` but it is _not_ what we ship with [what4 solvers](https://github.com/GaloisInc/what4-solvers/releases/tag/snapshot-20240212) as that is `v4.8.14`.  Adding a ticket to ensure we do `:check-docstring` against multiple z3 versions. We should also consider adding support to be able to set z3 version.

mccleeary-galois:
> Caught this on this PR, but the ML-KEM proofs currently require `z3 v4.13+`. This is what is installed by default with `brew install cryptol` but it is _not_ what we ship with [what4 solvers](https://github.com/GaloisInc/what4-solvers/releases/tag/snapshot-20240212) as that is `v4.8.14`. Adding a ticket to ensure we do `:check-docstring` against multiple z3 versions. We should also consider adding support to be able to set z3 version.

Going to cut some tickets for CI related tasks from this and also try and figure out which properties are actually causing the problem.

mccleeary-galois:
> For the second part, I am honestly kind of confused. I thought the endianness was wrong but I tried rewriting it multiple times to make it look more like the spec and it seems correct. The KATs pass and all the internal properties pass. If anyone discerns something weird I'm doing with bit ordering I would love feedback.

Looking at the spec I agree that this is not super clear on endianness here. I am happy with how this looks right now.

mccleeary-galois (approved):
LGTM, opened up a separate issue https://github.com/GaloisInc/cryptol-specs/issues/166 for the issue uncovered during review of this.

marsella:
I had to make a number of picky little changes that overlapped with the encoding pr #160. Nothing very interesting.

In service of #166 I switched a few properties to `exhaust` instead of `prove`, so I believe this now works with both z3 versions (the latest 4.14 and the shipped version 4.8.14)

PR #167, opened 2024-10-31
Delete broken 1.0 version of SphincsPlus.

mccleeary-galois:
Deleting old version of SphincsPlus that is broken and has been replaced with 3.1 version of spec.


marsella (approved):
Looks good, thank you!

PR #169, opened 2024-11-04
Add one-time signature scheme from XMSS

### LINKED ISSUE #162: Implement WOTS+ signature scheme
In order to implement XMSS, we need [the WOTS+ primitive signature scheme](https://datatracker.ietf.org/doc/html/rfc8391#section-3.1).

- [x] Define a functor with the parameters (3.1.1)
- [x] Define the chaining function (3.1.2)
- [x] Define the private key type and add warnings about its creation (3.1.3)
  - [ ] Add docs about pseudorandom key generation from a seed
- [x] Define the public key type and the derivation function (3.1.4)
- [x] Define the signature type and sign function (3.1.5)
- [x] Define signature verification (3.1.6)
- [x] Add some basic properties of signing functions (sign + verify works)
- [x] Look for KATs? I suspect they don't exist but it's worth checking.

marsella:
Closes #162.

This adds [the WOTS+ signature scheme](https://datatracker.ietf.org/doc/html/rfc8391#section-3.1) that's used in XMSS. It's very slow! About 50s for each public function (keygen, sign, verify). 

This will be unusably slow in the context of XMSS -- e.g. XMSS key generation calls WOTS+ key generation `2^h` times, where $h \in [10, 16, 20]$ -- for the largest parameter, this would take 1048576 minutes, or roughly 728 days, without accounting for the additional hashing that happens in XMSS.

Anyway I optimized `base_w` a bit but I think the main cost is coming from the recursive hashing; if we want this to be faster we will probably need to go look at SHA2 (the required parameter sets) and SHA3 (the optional parameter sets). In general, the goal of these specs is not to be performant -- it's just to look exactly like the written NIST specs -- but this appears to be getting so slow that we won't even be able to check basic properties of XMSS.

Another point of interest: this takes longer than usual to load (~10 - 30s), I suspect because the type constraints have a lot of computation in them.

Aside from performance, I think this is fairly straightforwardly representative of the spec. I'll add some inline questions and notes.

### DIFF
--- a/Primitive/Asymmetric/Signature/XMSS/Instantiations/WOTS_256.cry
+++ b/Primitive/Asymmetric/Signature/XMSS/Instantiations/WOTS_256.cry
@@ -1,3 +1,25 @@
+/**
+ * An instantiation of the WOTS+ one-time signature scheme.
+ *
+ * These parameters and naming are drawn from [SP-800-208], Section 5 (Table 9
+ * and Section 5.1). The same parameters are also defined in [RFC-8391]
+ * Section 5.1 (SHA2 with `n = 32`).
+ *
+ * References:
+ * [SP-800-208]: David A. Cooper, Daniel C. Apon, Quynh H. Dang, Michael S.
+ *     Davidson, Morris J. Dworkin, and Carl A. Miller. Recommendation for
+ *     Stateful Hash-Based Signature Schemes. (National Institute of Standards
+ *     and Technology, Gaithersburg, MD), NIST Special Publication (SP) NIST
+ *     SP 800-208. October 2020.
+ *     @see https://doi.org/10.6028/NIST.SP.800-208
+ * [RFC-8391]: Andreas Huelsing, Denis Butin, Stefan-Lukas Gazdag, Joost
+ *     Rijneveld, and Aziz Mohaisen. XMSS: eXtended Merkle Signature Scheme.
+ *     Internet Requests for Comments (RFC) 8391. May 2018.
+ *     @see https://datatracker.ietf.org/doc/rfc8391
+ *
+ * @copyright Galois, Inc.
+ * @author Marcella Hastings <marcella@galois.com>
+ */
 module Primitive::Asymmetric::Signature::XMSS::Instantiations::WOTS_256 =
     Primitive::Asymmetric::Signature::XMSS::Wots where
         import Primitive::Keyless::Hash::SHA2::Instantiations::SHA256 as SHA256
--- a/Primitive/Asymmetric/Signature/XMSS/Instantiations/WOTS_256.cry
+++ b/Primitive/Asymmetric/Signature/XMSS/Instantiations/WOTS_256.cry
@@ -1,0 +1,11 @@
+module Primitive::Asymmetric::Signature::XMSS::Instantiations::WOTS_256 =
+    Primitive::Asymmetric::Signature::XMSS::Wots where
+        import Primitive::Keyless::Hash::SHA2::Instantiations::SHA256 as SHA256
+
+        type n = 32
+        type w = 16
+
+        F KEY M = split (SHA256::hash (join ((zero : [32][8]) # KEY # M)))
+        PRF KEY M = split (SHA256::hash (join ((zero : [31][8]) # [(3 : [8])] # KEY # M)))
+
+
--- a/Primitive/Asymmetric/Signature/XMSS/Wots.cry
+++ b/Primitive/Asymmetric/Signature/XMSS/Wots.cry
@@ -46,6 +46,26 @@ parameter
     PRF : [n][8] -> [32][8] -> [n][8]
 
     /**
+     * A length value (fixed with respect to `n` and `w`).
+     * [RFC-8391] Section 3.1.1.
+     */
+    type len_1 = (8 * n) /^ (lg2 w)
+
+    /**
+     * A length value (fixed with respect to `n` and `w`).
+     * [RFC-8391] Section 3.1.1.
+     */
+
+    type len_2 = lg2 (len_1 * (w - 1)) / lg2 w + 1
+
+    /**
+     * The number of `n`-byte string elements in a WOTS+ private key, public key,
+     * and signature.
+     * [RFC-8391] Section 3.1.1.
+     */
+    type len = len_1 + len_2
+
+    /**
      * The `len` parameter fits in 32 bits.
      *
      * This implicit fact about our inputs is hard for Cryptol to infer, so we
@@ -59,8 +79,7 @@ parameter
      * This can also be observed from the computed `len` values from the valid
      * parameter sets in Section 5.2.
      */
-    type constraint (32 >= width ((8 * n) /^ (lg2 w)
-        + lg2 (((8 * n) /^ (lg2 w)) * (w - 1)) / lg2 w + 1))
+    type constraint (32 >= width len)
 
 /**
  * A byte is a sequence of 8 bits using big-endian representation.
@@ -74,22 +93,6 @@ type Byte = [8]
  */
 type Word = [32]
 
-/**
- * A length value.
- * [RFC-8391] Section 3.1.1.
- */
-type len_1 = (8 * n) /^ (lg2 w)
-/**
- * A length value.
- * [RFC-8391] Section 3.1.1.
- */
-type len_2 = lg2 (len_1 * (w - 1)) / lg2 w + 1
-/**
- * The number of `n`-byte string elements in a WOTS+ private key, public key,
- * and signature.
- * [RFC-8391] Section 3.1.1.
- */
-type len = len_1 + len_2
 
 /**
  * Address scheme for randomizing hash function calls in the OTS scheme.
@@ -191,8 +194,6 @@ private
         BM = PRF SEED ADRS'''
 
         tmp' = F KEY (tmp ^ BM)
-
-private
 
     /**
      * Convert a byte string into a set of base-`w` numbers.
--- a/Primitive/Asymmetric/Signature/XMSS/Wots.cry
+++ b/Primitive/Asymmetric/Signature/XMSS/Wots.cry
@@ -7,9 +7,15 @@
  * Warning: If a private key is used to sign two different messages, the scheme
  * becomes insecure! Cryptol cannot protect against this failure mode!
  *
+ * Instantiation warnings
+ * - The WOTS+ signature scheme includes a checksum. Per [RFC-8391], the
+ *   checksum can fit in 32 bits _when instantiated with an approved parameter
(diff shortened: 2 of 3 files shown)


marsella:
Here's some initial benchmarks I ran. `sign` has a couple of measurements because it turns out all-zero input circumvents most of the function in a way that is not representative of real data.

```
> :time genPK zero zero zero
Measuring for 300 seconds
Avg time: 51.98 s    Avg CPU time: 52.03 s    Avg cycles: 1247554038
> 
> :time WOTS_sign zero zero zero zero
Measuring for 300 seconds
Avg time: 2.598 ms    Avg CPU time: 2.600 ms    Avg cycles: 62349
> :time WOTS_sign (zero # ADRS) zero SEED ADRS 
Measuring for 300 seconds
Avg time: 22.73 s    Avg CPU time: 22.73 s    Avg cycles: 545567978
>
> :time WOTS_pkFromSig  zero zero zero zero
Measuring for 300 seconds
Avg time: 51.43 s    Avg CPU time: 51.45 s    Avg cycles: 1234365157
```

Primitive/Asymmetric/Signature/XMSS/Instantiations/WOTS_256.cry
```diff
@@ -0,0 +1,11 @@
+module Primitive::Asymmetric::Signature::XMSS::Instantiations::WOTS_256 =
```
marsella, line 1: todo: I need to document where these parameters came from. They're in section 5. Also add copyright notice.

Primitive/Asymmetric/Signature/XMSS/Instantiations/WOTS_256.cry
```diff
@@ -0,0 +1,11 @@
...
+    Primitive::Asymmetric::Signature::XMSS::Wots where
+        import Primitive::Keyless::Hash::SHA2::Instantiations::SHA256 as SHA256
+
+        type n = 32
+        type w = 16
+
+        F KEY M = split (SHA256::hash (join ((zero : [32][8]) # KEY # M)))
+        PRF KEY M = split (SHA256::hash (join ((zero : [31][8]) # [(3 : [8])] # KEY # M)))
```
marsella, line 9: nb: I think I need to add a `hashBytes` function to all the hashes because this is starting to be a really typical pattern (`split (hash (join x))`).

Primitive/Asymmetric/Signature/XMSS/Wots.cry
```diff
@@ -0,0 +1,376 @@
...
+     * A length value (fixed with respect to `n` and `w`).
+     * [RFC-8391] Section 3.1.1.
+     */
+
+    type len_2 = lg2 (len_1 * (w - 1)) / lg2 w + 1
+
+    /**
+     * This type is used in Algorithms 5 and 6; it's the padded length of a
+     * checksum. The checksum has an upper bound of 32 bits, so the padding
+     * must be at least 32 bits.
```
marsella, line 74: todo: I'm still kind of confused about this; the spec doesn't clarify what to do if the padding is shorter than the checksum, so I assume it needs to be at least as long. However in my instantiation it was shorter (only 2 bytes). Maybe I filled in the instantiation incorrectly? 

Anyway I took out this type constraint but I need to figure out what's going on here.

Primitive/Asymmetric/Signature/XMSS/Wots.cry
```diff
@@ -0,0 +1,376 @@
...
+
+/**
+ * Convert a non-negative integer to a binary representation in big-endian
+ * byte order.
+ * [RFC-8391] Section 2.4.
+ */
+toByte : {y, m} (fin y, y > 0) => [m] -> [y]Byte
+toByte x
+    | m <= 8 * y => split (zext x)
+    | m > 8 * y => split (take x)
```
marsella, line 115: todo: this relates to the comment above -- the spec doesn't say anything about the `m > 8y` case. I need to try to figure out whether it should be here or not.

Primitive/Asymmetric/Signature/XMSS/Wots.cry
```diff
@@ -0,0 +1,376 @@
...
+ * 4. Type (fixed at 0)
+ * 5. OTS address
+ * 6. Chain address
+ * 7. Hash address
+ * 8. keyAndMask
+ *
+ * Note that this isn't defined in terms of the `Word` type because it's
+ * operated on by things that require arrays of bytes.
+ */
+type OTSHashAddress = [8 * 4]Byte
```
marsella, line 139: note: open to suggestions on whether to represent this differently. E.g. it could be a record with named fields. This "looks" the most like the ascii representation of the type in Section 2.5 but I'm not sure that's the obvious best way to represent it.

Primitive/Asymmetric/Signature/XMSS/Wots.cry
```diff
@@ -0,0 +1,376 @@
...
+ * A private, or secret, key in WOTS+ is a length `len` array of `n`-byte
+ * strings.
+ *
+ * It represents the start nodes in a set of hash chains.
+ *
+ * ⚠️ Warning ⚠️: A private key MUST be selected randomly from the uniform
+ * distribution or selected using a cryptographically secure pseudorandom
+ * process! Cryptol cannot verify that a `PrivateKey` was chosen suitably!
+ * Implementors must independently audit private key generation!
+ *
```
marsella, line 151: todo: There's a bit about pseudorandom key generation in the spec; it calls out to a separate doc with a pseudorandom method. Consider adding that function to get a key from a seed or at least describing it in more detail here.

Primitive/Asymmetric/Signature/XMSS/Wots.cry
```diff
@@ -0,0 +1,376 @@
...
+ * It represents the end nodes i a set of length-`w` hash chains, where the
+ * start nodes are defined in the corresponding `PrivateKey`.
+ */
+type PublicKey = [len][n]Byte
+
+/**
+ * Generate a WOTS+ public key from a private key.
+ */
+genPK : PrivateKey -> OTSHashAddress -> [n]Byte -> PublicKey
+genPK sk ADRS SEED = pk where
```
marsella, line 168: question (naming): In the spec this is `WOTS_genPK`. Preference on using the prefix (here and in the other functions) vs relying on the namespace to "fill in" the prefix in the calling spots (e.g. call it `genPK` and invoke it using `WOTS::genPK`).

Primitive/Asymmetric/Signature/XMSS/Wots.cry
```diff
@@ -0,0 +1,376 @@
...
+        basew = take`{consumed} [ zext ((total >> bits) && (`w - 1))
+            // Steps 1-5. Initialize helper variables (and update them for each
+            // round). The first element is dropped because the first variable
+            // update happens before computing any output values.
+            | (total, in, bits) <- drop`{1} (iterate update_helpers (0, 0, 0))]
+
+    /**
+     * As a point of interest, we also implement the `base_w` function using
+     * built-in Cryptol functions, just regrouping the bits into the desired
+     * arrangement.
```
marsella, line 270: todo: update docs, explain speed improvement (1.046 ms -> 197.8 μs).

Primitive/Asymmetric/Signature/XMSS/Wots.cry
marsella: I checked [the reference implementation](https://github.com/XMSS/xmss-reference/blob/master/utils.c#L6) and the answer is: we truncate the checksum. Removed this comment as it is no longer correct, and since this doesn't need a type constraint I took it out of the interface.

Primitive/Asymmetric/Signature/XMSS/Wots.cry
marsella: As noted above, I checked [the reference implementation](https://github.com/XMSS/xmss-reference/blob/master/utils.c#L6) and fixed this. Actually this was the correct behavior but I made it a 1-liner instead, and added an explanation and a link.

kiniry: Doesn't this indicate that we should give feedback to NIST about underspecification problems in the standard?

Primitive/Asymmetric/Signature/XMSS/Wots.cry
```diff
@@ -0,0 +1,388 @@
...
+     */
+    type n : #
+    type constraint (fin n)
+
+    /**
+     * The Winternitz parameter. This must be either 4 or 16.
+     * [RFC-8391] Section 3.1.1.
+     */
+    type w : #
+    type constraint (fin w, w % 4 == 0, w % 12 == 4, w <= 16)
```
mccleeary-galois, line 43: I feel like this should be something easier to say as it is quite common to have a parameter be from a set...

mccleeary-galois: Not seeing any better options here https://github.com/GaloisInc/cryptol/blob/876e97a745d157cf0048551ea7f4d2d3f1bca35a/src/Cryptol/TypeCheck/TCon.hs#L40

Primitive/Asymmetric/Signature/XMSS/Wots.cry
mccleeary-galois: I think relying on the namespace makes sense here,

`WOTS::WOTS_genPK` feels redundant

mccleeary-galois (approved):
LGTM for this part of XMSS

Primitive/Asymmetric/Signature/XMSS/Wots.cry
marsella: If that's an option I'd be happy to write up a note about it!

Primitive/Asymmetric/Signature/XMSS/Wots.cry
marsella: I'll add a comment walking through it.

Primitive/Asymmetric/Signature/XMSS/Wots.cry
marsella: Going to keep this as-is for now; if we find that it would be better represented another way later I'll change it.

Primitive/Asymmetric/Signature/XMSS/Wots.cry
marsella: Added docs.

Primitive/Asymmetric/Signature/XMSS/Wots.cry
marsella: Took out the prefixes and added a note in the docs with the spec name (in case anyone's ctrl-F'ing through the repo to find it).

PR #170, opened 2024-11-06
ML-KEM: add input validation functions

### LINKED ISSUE #149: Add input validation functions and docs to ML-KEM
The current implementation of ML-KEM does not quite align with [the spec](https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.203.pdf). The spec defines "internal" functions that run the actual computations, and "external" functions that require input validation and sample / check randomness in a specific way. We have functions whose names match the "external" functions but implementation and docs match the "internal" ones. We also don't have any of the required input validation implemented.

This is for the `Encaps`, `Decaps`, and `Keygen` functions in ML-KEM.

- [ ] Rename existing functions to `_internal` and update docs accordingly
- [ ] Make non-suffix "external" functions that (a) throw an error if the randomness is not suitable and (b) document the required input validation
- [ ] Make (public) input validation functions
	- [ ] Key pair check in 7.1
	- [ ] Encapsulation key check in 7.2
	- [ ] Decapsulation input check in 7.3
- [ ] Make sure documentation across the board is clear about failure cases if you don't validate input. Also, add big warnings about randomness generation that Cryptol cannot do itself.

marsella:
Closes #149.

This rearranges the public API:
- Renames and moves to private the existing functions, to better match the spec
- Adds a new public API (and replaces all calls)
- Adds input validation functions for the new public API
- Adds warnings about use of the updated functions.


Primitive/Asymmetric/Cipher/ML_KEM/Specification.cry
```diff
@@ -1,6 +1,21 @@
 /**
  * ML-KEM cipher with a fast NTT implementation.
  *
+ * This implements the final version of ML-KEM as standardized in [FIPS-203].
+ * The executable spec differs from the spec in several important ways:
```
marsella, line 5: See also the implementation concerns in Section 3.3.

Primitive/Asymmetric/Cipher/ML_KEM/Specification.cry
```diff
@@ -1164,77 +1179,262 @@ private submodule K_PKE where
...
+ * cannot generate randomness).
  *
- * The randomness here is passed explicitly as a parameter and is not checked
- * for validity, so this actually corresponds to
- * [FIPS-203] Section 6.1 Algorithm 16 (`ML_KEM.KeyGen_internal`).
+ * The decapsulation key produced from this function MUST remain private. The
+ * seed `(d, z)` produced (or in this case, passed as input) is sensitive data
+ * and MUST be treated with the same safeguards as the decapsulation key.
+ */
+KeyGen : Option ([32]Byte) -> Option ([32]Byte)
```
marsella, line 1303: question: here and in `Encaps`, the functions are supposed to generate randomness, where randomness can be either a byte vector or NULL (see Section 2.4, exception 1). I've tried to represent this by taking an `Option`al parameter. It does make calling these functions way clunkier. Thoughts on whether there's a better alternate way?

Primitive/Asymmetric/Cipher/ML_KEM/Tests/kat.awk
marsella, line 1: Full disclosure: I didn't actually re-test this, but it looks like the right API. I can go try it out if reviewers request.

Primitive/Asymmetric/Cipher/ML_KEM/Specification.cry
mccleeary-galois: Option makes sense to me for NULL handling... We could make a helper that just calls this and defaults this this to `None` if we another endpoint?

Primitive/Asymmetric/Cipher/ML_KEM/Tests/kat.awk
mccleeary-galois: Yeah we probably should... We need to at least change the `:load` line to
`:load Primitive/Asymmetric/Cipher/ML_KEM/Instantiations/ML_KEM.cry`

mccleeary-galois (approved):
LGTM, lets fix the `:load` command in the `.awk` file though.

Primitive/Asymmetric/Cipher/ML_KEM/Specification.cry
marsella: Sounds good. I'll probably leave as-is for now but if we get feedback from clients / users I'd be happy to add.

PR #190, opened 2024-11-18
ML-DSA: Define parameters and constants

### LINKED ISSUE #180: ML-DSA: Make top-level file
We need to make a final version of [ML-DSA](https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.204.pdf). I think the best option is to start from scratch in a new file and copy over useful components from the initial public draft version that already exists.

The first thing to add is the parameters and types that are defined across the rest of the spec. Section 4, table 1 defines all the parameters for the library. Some are fixed across all parameter sets and can be defined in this module. Others vary across different parameter sets and need to be defined in the [anonymous interface / parameter block](https://galoisinc.github.io/cryptol/master/Modules.html#anonymous-interface-modules). Each parameter and type added should have documentation explaining what it is (can be drawn from the first column in the table) and a reference to the place it's defined in the spec. 

We should also try to define type constraints that limit the parameters to only the allowable values defined in the spec. Some of these are not very obvious at first glance; we might have to make judgement calls about which ones are straightforwardly definable and which are too convoluted to be useful.

I think we don't need to try to define the analysis parameters at this point -- that's the last 3 rows of the table (challenge entropy, repetitions, and claimed security strength).

- [x] Make a new module `Primitive/Asymmetric/Signature/MLDSA/Specification.cry` with a short high-level description of the algorithm in the module docs.
- [x] Write out the correct reference for the NIST spec in the module docs.
- [x] Define the parameters for the system based on Section 4, Table 1.
- [x] Define the constants for the system based on Section 4, Table 1.

marsella:
Closes #180.

Looking for feedback on whether the docs are reasonable. I tried to add a little more context that was provided inline in the spec, but if it gets too much into weeds I will remove or reword.

### DIFF
--- a/Primitive/Asymmetric/Signature/ML_DSA/Specification.cry
+++ b/Primitive/Asymmetric/Signature/ML_DSA/Specification.cry
@@ -60,13 +60,12 @@ parameter
      * [FIPS-204] Section 4, Table 1.
      *
      * The type constraint is drawn from [FIPS-204] Section 7.4. The primary
-     * constraint we wish to encode is that `2 * γ2` divides `q - 1` (the floor
-     * and ceiling division operations only match when the division is exact).
-     * In order to express this, we must also require that γ2 is finite and
-     * non-zero.
+     * constraint we wish to encode is that `2 * γ2` divides `q - 1` (that is,
+     * if you divide `q - 1` by `2 * γ2`, the remainder is 0).
+     * In order to express this, we must also require that γ2 is non-zero.
      */
     type γ2 : #
-    type constraint (fin γ2, 2 * γ2 > 0, (q - 1) / (2 * γ2) == (q - 1) /^ (2 * γ2))
+    type constraint (2 * γ2 > 0, (q - 1) % (2 * γ2) == 0)
 
     /**
      * Dimensions of the matrix `A`, which is part of the public key.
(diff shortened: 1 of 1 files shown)


Primitive/Asymmetric/Signature/ML_DSA/Specification.cry
```diff
@@ -0,0 +1,112 @@
...
+    type γ1 : #
+    type constraint (2 ^^ (lg2 γ1) == γ1)
+
+    /**
+     * Low-order rounding range; this defines how to round the signer's
+     * commitment `w`.
+     * [FIPS-204] Section 4, Table 1.
+     *
+     * The type constraint is drawn from [FIPS-204] Section 7.4. The primary
+     * constraint we wish to encode is that `2 * γ2` divides `q - 1` (that is,
```
mccleeary-galois, line 63: The parenthesized part of this probably isn't necessary, but other than that I don't mind the extra information about where the constraints are coming from.

mccleeary-galois (approved):
Looks good overall, I appreciate knowing where the constraints actually came from as they are often not explicit at the beginning of the spec.

Primitive/Asymmetric/Signature/ML_DSA/Specification.cry
marsella: Oh gosh yes. The type constraint used to be way more complicated but then I remembered that remainders exist... I'll take this out.

PR #195, opened 2024-11-21
ML-DSA: Add rejection sampling functions

### LINKED ISSUE #182: ML-DSA: Add NTT rejection sampling functions
In order to sample randomly from the NTT rings `Tq` and `R` for [ML-DSA](https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.204.pdf), we need to support some rejection sampling functions. These use the SHAKE extendable-output function to get a stream of pseudorandom bytes from a string, then checks each chunk of the stream to see if it's a suitable value, throwing it away if not. 

This will be somewhat different from the version in the IPD for a few reasons.
- The SHAKE API has changed - there's now an `xof` function that allows infinite output. Previously we had sampled a large, finite stream.
- The boundary between stream parsing and rejection sampling seems to have moved. In the IPD implementation, the `coeffFrom*Byte` functions take infinite streams, but in the final spec they each take finite input. We want to match the version in the spec. (Or maybe this is because we didn't previously have enum support. We do now! So we'll use it.)

One notable difference from the spec is that we'll use an infinite output stream from the XOF instead of using the init/absorb/squeeze API. Just make a note that they're equivalent in the docs and refer to section 3.7.

All these functions can be private. They're in one issue because they seem logically grouped to me, but they live in different parts of the spec; try to physically arrange them in the same order that the spec has them. As always, make sure all new functions have appropriate documentation, citations to specific locations in the spec, and match (in naming and steps) the spec as closely as possible.

- [x] Define the NTT-related types we need for this (`Tq` and `R`) -- see sections 2.3 (definition) and 2.4.1 (representation). They're length-256 arrays of elements in `Z` or `Zq`, but it's useful to define them separately.
- [x] Define the hash functions `H` and `G`, using the new `xof` API (Section 3.7).
- [x] Implement `CoeffFromThreeBytes` and `CoeffFromHalfByte`. These should return `Option`s. I think the correct spec-adherent implementation will be a subset of the versions in the IPD.
- [x] Implement `RejNTTPoly`
- [x] Implement `RejBoundedPoly`. The control flow on this is a little awkward to implement in Cryptol; see the ML-KEM `SampleNTT` function for one way to do it.
- [x] Add docs about equivalence of squeeze vs infinite stream

marsella:
Closes #182.

Adds a few functions for rejection sampling. 

I compared some of these to the versions in the IPD. The hash functions are the same. The `CoeffFrom...` functions are the same. `RejBoundedPoly` matches. The seed for `RejNTTPoly` is a different size and I tried to compare them, but it didn't work. I feel pretty confident about my implementation, though, so I'm willing to wait to see if we get an error at a higher level.

### DIFF
--- a/Primitive/Asymmetric/Signature/ML_DSA/Specification.cry
+++ b/Primitive/Asymmetric/Signature/ML_DSA/Specification.cry
@@ -25,7 +25,7 @@ import Primitive::Keyless::Hash::SHAKE::SHAKE256 as SHAKE256
 type Byte = [8]
 
 /**
- * Ring defined as the product of 255 elements in `Z q` used for NTT.
+ * Ring defined as the product of 256 elements in `Z q`, used for NTT.
  * [FIPS-204] Section 2.3 and Section 2.4.1.
  */
 type Tq = [256](Z q)
@@ -49,7 +49,7 @@ type R = [256]Integer
  * in the same section.
  */
 H : {l, m} (fin m) => [m][8] -> [l][8]
-H str = split (SHAKE256::xof`{8 * l} (join str))
+H str = SHAKE256::xofBytes`{8 * l} str
 
 /**
  * Wrapper function around SHAKE128, specifying the length `l` in bytes.
@@ -61,7 +61,7 @@ H str = split (SHAKE256::xof`{8 * l} (join str))
  * in the same section.
  */
 G : {l, m} (fin m) => [m][8] -> [l][8]
-G str = split (SHAKE128::xof`{8 * l} (join str))
+G str = SHAKE128::xofBytes`{8 * l} str
 
 parameter
     /**
@@ -212,6 +212,10 @@ RejNTTPoly ρ = a_hat where
         // Step 7 - 9. The recursive call here replaces the `while` loop.
         a_hat' = case a_hat_j of
             Some aj -> [aj] # (sample ctx)
+            // In the spec, the sample `a_hat_j` is always added to the list,
+            // and `j` is only increased if the sample was not rejected (so a
+            // rejected value is overwritten in the next iteration). Here,
+            // we only add `a_hat_j` if it's valid.
             None -> sample ctx
 
 /**
--- a/Primitive/Asymmetric/Signature/ML_DSA/Specification.cry
+++ b/Primitive/Asymmetric/Signature/ML_DSA/Specification.cry
@@ -23,6 +23,21 @@ import Primitive::Keyless::Hash::SHAKE::SHAKE128 as SHAKE128
 import Primitive::Keyless::Hash::SHAKE::SHAKE256 as SHAKE256
 
 /**
+ * Ring defined as the product of 255 elements in `Z q` used for NTT.
+ * [FIPS-204] Section 2.3 and Section 2.4.1.
+ */
+type Tq = [256](Z q)
+
+/**
+ * Ring of single-variable polynomials over the integers mod `X^256 + 1`.
+ * [FIPS-204] Section 2.3 and Section 2.4.1.
+ *
+ * The `i`th element of this list represents the coefficient for the degree-`i`
+ * term.
+ */
+type R = [256]Integer
+
+/**
  * Wrapper function around SHAKE256, specifying the length `l` in bytes.
  * [FIPS-204] Section 3.7.
  *
--- a/Primitive/Asymmetric/Signature/ML_DSA/Specification.cry
+++ b/Primitive/Asymmetric/Signature/ML_DSA/Specification.cry
@@ -188,3 +188,28 @@ CoeffFromHalfByte b =
     else
         if (`η == 4) && (b < 9) then Some (4 - toInteger b)
         else None
+
+/**
+ * Sample a polynomial in the ring `Tq`.
+ * [FIPS-204] Section 7.3, Algorithm 30.
+ */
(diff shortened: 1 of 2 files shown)


Primitive/Asymmetric/Signature/ML_DSA/Specification.cry
```diff
@@ -19,6 +19,50 @@
  */
 module Primitive::Asymmetric::Signature::ML_DSA::Specification where
 
+import Primitive::Keyless::Hash::SHAKE::SHAKE128 as SHAKE128
+import Primitive::Keyless::Hash::SHAKE::SHAKE256 as SHAKE256
+
+type Byte = [8]
+
+/**
+ * Ring defined as the product of 255 elements in `Z q` used for NTT.
```
mccleeary-galois, line 28: The array is still 256 for this correct since it is storing X^^0 as well? I get that this only goes up to X^^255 is that what you are implying by this being 255?

Primitive/Asymmetric/Signature/ML_DSA/Specification.cry
```diff
@@ -109,3 +153,113 @@ type d = 13
...
+        // Step 5. This pops the first 3 bytes off the pseudorandom stream.
+        (s, ctx) = splitAt`{3} GSqueeze
+
+        // Step 6.
+        a_hat_j = CoeffFromThreeBytes (s@0) (s@1) (s@2)
+
+        // Step 7 - 9. The recursive call here replaces the `while` loop.
+        a_hat' = case a_hat_j of
+            Some aj -> [aj] # (sample ctx)
+            None -> sample ctx
```
mccleeary-galois, line 215: not sure if it is worth mentioning that the spec just overwrites this part of the "stream" on this case hence why it is safe to just skip writing it

Primitive/Asymmetric/Signature/ML_DSA/Specification.cry
```diff
@@ -110,3 +154,113 @@ type d = 13
...
+    // Step 2 - 3.
+    ctx0 = G ρ
+
+    // Step 4, 11. The `take` here replaces the loop condition.
+    a_hat = take`{256} (sample ctx0)
+
+    sample : [inf][8] -> [inf](Z q)
+    sample GSqueeze = a_hat' where
+        // Step 5. This pops the first 3 bytes off the pseudorandom stream.
+        (s, ctx) = splitAt`{3} GSqueeze
```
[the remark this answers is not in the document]
mccleeary-galois: I was able to understand it as is.

Primitive/Asymmetric/Signature/ML_DSA/Specification.cry
```diff
@@ -109,3 +153,113 @@ type d = 13
...
+
+        // Step 6 - 8. We use Cryptol-native functions instead of dividing
+        // and modding `z`. See `TakeAndDropAreDivAndMod` for the equivalence.
+        z0 = CoeffFromHalfByte (drop`{4} z)
+        z1 = CoeffFromHalfByte (take`{4} z)
+
+        // Step 8 - 15. The recursive calls replace the `while` loop.
+        // In order to make the types work, we have to mash the two conditions
+        // together and make exactly one recursive call.
+        a' = case z0 of
```
mccleeary-galois, line 242: nit: I might prefer a single case statement here ie

```
case (z0,z1) of
    (Some z0', Some z1') -> [z0', z1'] # (sample ctx)
    (Some z0', None) -> [z0'] # (sample ctx)
    (None, Some z1') -> [z1'] # (sample ctx)
    (None, None) -> sample ctx

File: Primitive/Asymmetric/Signature/ML_DSA/Specification.cry [hunk 5]
```diff
@@ -110,3 +154,113 @@ type d = 13
...
+        else None
+
+/**
+ * Sample a polynomial in the ring `Tq`.
+ * [FIPS-204] Section 7.3, Algorithm 30.
+ */
+RejNTTPoly : [32]Byte -> Tq
+RejNTTPoly ρ = a_hat where
+    // Step 2 - 3.
+    ctx0 = G ρ
```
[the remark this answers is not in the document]
mccleeary-galois: maybe a quick note is worth a drop, but I don't think it is necessary

mccleeary-galois (approved):
LGTM

Primitive/Asymmetric/Signature/ML_DSA/Specification.cry
marsella: Oh you're right, that was just a mistake on my part. The definition on page 6 is  $$\prod_{j=0}^{255} \mathbb Z_q$$ and I misparsed it.

Primitive/Asymmetric/Signature/ML_DSA/Specification.cry
marsella: I would too! Sadly Cryptol doesn't support nested case statements right now. I made an issue to request the feature (https://github.com/GaloisInc/cryptol/issues/1779).

Primitive/Asymmetric/Signature/ML_DSA/Specification.cry
marsella: Good call, this was confusing when I came back to it. Added some docs.

PR #200, opened 2024-12-03
ML-DSA: Add conversion functions

### LINKED ISSUE #181: ML-DSA: Add conversion functions
There are a bunch of conversion functions in section 7 of [the ML-DSA spec](https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.204.pdf). These seem to be in mostly good shape in the IPD version we have implemented, but the documentation is out of date and doesn't follow our current format for references.

There are examples of good documentation in ML-KEM, among other places, but the basic goals are:
- Top-level doc comments should be formatted like: 1 line explaining what the function does (often pulled from the spec directly); citation for where it is in the spec (`[spec-ref] Section #, Algorithm #.`), then any additional notes or comments (e.g. does it differ from the spec? how and why? Note if an input to the function is defined in the type signature instead. Note anything otherwise confusing.)
- If it's a property, the top-level comment should include a `repl` block that proves, checks, or exhausts the property.
- Inline comments should mark which steps are handled by each line of code, especially in cases where it's not obvious or is in a slightly different order from the spec.
- Inline comments can explain any differences from the spec
- Block comments use `/**` format; inline comments use `//`. There's a space between the comment marker and the text (e.g. `// <--There's a space right there.`).

Also, all of these functions are internal so they should be `private`.

- [x] Copy and update `IntegerToBits` function
- [x] Copy and update `BitsToInteger` function
- [x] Add property showing the integer/bits conversion functions are each others inverse.
- [x] Write `IntegerToBytes` function (this doesn't seem to exist in the ipd version but should be similar to `IntegerToBits`).
- [x] Bring `BitsToBytes` and `BytesToBits` functions in from `ML-KEM` -- the spec versions are identical up to naming, so it's probably easiest to copy the code from ML-KEM and update the docs and references appropriately.

marsella:
Closes #181.

This adds the conversion functions between bits, bytes, and integers from the spec.

### DIFF
--- a/Primitive/Asymmetric/Signature/ML_DSA/Specification.cry
+++ b/Primitive/Asymmetric/Signature/ML_DSA/Specification.cry
@@ -202,6 +202,19 @@ BitsToIntegersInverts : {α} (fin α, α > 0) => [α] -> Bit
 property BitsToIntegersInverts x = IntegerToBits (BitsToInteger x) == x
 
 /**
+ * Compute a base-256 representation of `x mod 256^α` using little-endian byte
+ * order.
+ * [FIPS-204] Section 7.1, Algorithm 11.
+ */
+IntegerToBytes : {α} (fin α, α > 0) => Integer -> [α]Byte
+IntegerToBytes x = y where
+    // Step 3. Compute the value of `x'` at each iteration of the loop.
+    xs' = [x] # [x' / 256 | x' <- xs']
+
+    // Step 2, 4.
+    y = [fromInteger (x' % 256) | x' <- xs' | i <- [0..α - 1]]
+
+/**
  * Generate an element in the integers mod `q` or a failure indicator.
  * [FIPS-204] Section 7.1, Algorithm 14.
  */
--- a/Primitive/Asymmetric/Signature/ML_DSA/Specification.cry
+++ b/Primitive/Asymmetric/Signature/ML_DSA/Specification.cry
@@ -208,11 +208,11 @@ property BitsToIntegersInverts x = IntegerToBits (BitsToInteger x) == x
  */
 IntegerToBytes : {α} (fin α, α > 0) => Integer -> [α]Byte
 IntegerToBytes x = y where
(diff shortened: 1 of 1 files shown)


Primitive/Asymmetric/Signature/ML_DSA/Specification.cry
```diff
@@ -154,6 +154,136 @@ type d = 13
...
+ * order.
+ * [FIPS-204] Section 7.1, Algorithm 11.
+ */
+IntegerToBytes : {α} (fin α, α > 0) => Integer -> [α]Byte
+IntegerToBytes x = y where
+    // Step 3. Compute the value of `x'` at each iteration of the loop.
+    xs' = [x] # [x' / 256 | x' <- xs']
+
+    // Step 2, 4.
+    y = [fromInteger (x' % 256) | x' <- xs' | i <- [0..α - 1]]
```
mccleeary-galois, line 215: This looks like step 3 to me?

Primitive/Asymmetric/Signature/ML_DSA/Specification.cry
```diff
@@ -154,6 +154,136 @@ type d = 13
...
+
+/**
+ * Compute a base-256 representation of `x mod 256^α` using little-endian byte
+ * order.
+ * [FIPS-204] Section 7.1, Algorithm 11.
+ */
+IntegerToBytes : {α} (fin α, α > 0) => Integer -> [α]Byte
+IntegerToBytes x = y where
+    // Step 3. Compute the value of `x'` at each iteration of the loop.
+    xs' = [x] # [x' / 256 | x' <- xs']
```
mccleeary-galois, line 212: This looks like step 4 to me?

mccleeary-galois (approved):

Primitive/Asymmetric/Signature/ML_DSA/Specification.cry
marsella: Oops! Thanks.

PR #202, opened 2024-12-05
ML-DSA: Add rejection sampling functions

### LINKED ISSUE #188: ML-DSA: Implement sampling functions
Blocked by #182, #181, #183.
In addition to the NTT sampling functions in #182, there are several algorithms for generating algebraic objects pseudorandomly from a seed. 

For `SampleInBall`, the XOF API has changed since the IPD, so the function will need to be modified a bit. As elsewhere, we'll use the `xof` API and lazily pull off an infinite output instead of using the init/absorb/squeeze API in the text. Also, there is reportedly a difference in how many bits of `c~` are used here -- I'm not sure if that's at the call site or in the function itself.

I think the others should be fairly straightforward once all the dependent functions are implemented (see blockers).

- [x] Implement and document `SampleInBall`
- [x] Implement and document `ExpandA`
- [x] Implement and document `ExpandS`
- [x] Implement and document `ExpandMask` -- this reportedly uses `H` differently from in the IPD, so make sure it's used correctly.

marsella:
Closes #188 

This is not ready for review. TODO:
- [ ] Decide how to handle the conflicting type constraints in `bitunpack` and `expandMask`
- [ ] Remove copy of `BitUnpack` and rebase onto master once #201 is merged
- [ ] Review for clarity and docs.

### DIFF
--- a/Primitive/Asymmetric/Signature/ML_DSA/Specification.cry
+++ b/Primitive/Asymmetric/Signature/ML_DSA/Specification.cry
@@ -530,7 +530,7 @@ SampleInBall ρ = cFinal where
  * Sample a polynomial in the ring `Tq`.
  * [FIPS-204] Section 7.3, Algorithm 30.
  */
-RejNTTPoly : [32]Byte -> Tq
+RejNTTPoly : [34]Byte -> Tq
 RejNTTPoly ρ = a_hat where
     // Step 2 - 3.
     ctx0 = G ρ
--- a/Primitive/Asymmetric/Signature/ML_DSA/Specification.cry
+++ b/Primitive/Asymmetric/Signature/ML_DSA/Specification.cry
@@ -599,6 +599,16 @@ ExpandA ρ = A_hat where
         | s <- [0..ell - 1]]
         | r <- [0..k - 1]]
 
+
+/**
+ * Sample vectors in `R^ell` and `R^k`, each with polynomial coordinates whose
+ * coefficients are in the interval `[-η , η]`.
+ */
+ExpandS : [64]Byte -> ([ell]R, [k]R)
+ExpandS ρ = (s1, s2) where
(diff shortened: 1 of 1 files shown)


Primitive/Asymmetric/Signature/ML_DSA/Specification.cry
```diff
@@ -696,6 +744,56 @@ property TakeAndDropAreDivAndMod z = dropIsMod && takeIsDiv where
...
+    // However, the spec defines:
+    //     `c := 1 + width (γ1 - 1)`.
+    //
+    // Since we know that `γ1` is a power of 2, these two definitions are equivalent.
+    // Note that `width` is defined as `width x = lg2 (x + 1)`. We have:
+    // `BitUnpack` definition:
+    //     `width (γ1 - 1 + γ1) = lg2 (2 * γ1) = 1 + lg2 γ1`
+    // `ExpandMask` definition:
+    //     `1 + width (γ1 - 1) = 1 + lg2 γ1`
+    type c = width (2 * γ1 - 1)
```
marsella, line 790: todo: I realized I can put this equality in a type constraint instead of inline here (and I ran into it in another place, so better to have it at the top level anyway). It should always be true for the valid parameter sets and then I can put exactly what's in the spec right here.

marsella: I fixed this.

mccleeary-galois:
> Closes #188
> 
> @mccleeary-galois: Do you want me to write properties for the function descriptions? E.g. `ExpandMask` says "Samples [...] coefficients between −𝛾1 + 1 and 𝛾1.". I currently don't have a property that checks the output of these sampling functions to make sure that the values are in range, but I can add it pretty easily.

Let's add it if it isn't to much of an ask.

mccleeary-galois (approved):
LGTM, happy to have the properties in a follow on task.

marsella:
Thank you! Added #207.

PR #203, opened 2024-12-06
ML-DSA: add hint-related functions

### LINKED ISSUE #189: ML-DSA: Implement high & low bit and hint functions
We need to implement the various functions in [Section 7.4. ](https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.204.pdf)

The spec overloads these to apply both to individual values and to vectors (mapping the individual function to each element of the vector). Since  we can't overload names in cryptol, we'll use the same naming scheme that we used in ML-KEM: appending `_Vec` to the name for the vector functions.
That said, I think we might not need both versions of the function in every case. E.g. it looks like `Power2Round` is only used in the vector case; we might be fine to just define it for that case at the top level and have the Algorithm 35 be a sub-function (e.g. defined in the where clause). `Decompose` seems to only be used in the subsequent high/low bits and hint functions in this section, so maybe it doesn't need to be defined for vectors.

The IPD versions of these all look pretty good, so this should mainly be a docs update PR.

- [x] Audit use cases of these functions to determine which top-level function signatures we need.
- [x] Implement and update docs for `Power2Round`
- [x] Implement and update docs for `Decompose` 
- [x] Implement and update docs for `HighBits` and `LowBits`
- [x] Implement and update docs for `MakeHint` and `UseHint`

marsella:
Closes #189.

This adds a bunch of fairly straightforward functions for extracting high and low bits from integers and then making hints out of them.

There's a little bit of funkiness due to name overloading in the spec -- the functions are defined over single values but called over vectors of polynomials of values. So there's a bunch of iteration that is all implicit in the spec. I wanted to make the call sites for these functions as simple as possible, so I contained the single-value versions of the functions inside the main functions, but it does change the definitions a bit.

### DIFF
--- a/Primitive/Asymmetric/Signature/ML_DSA/Specification.cry
+++ b/Primitive/Asymmetric/Signature/ML_DSA/Specification.cry
@@ -681,3 +681,27 @@ private
             | r1s <- r1_vec
             | r0s <- r0_vec
             | rs <- r_vec]
+
+/**
+ * Returns `r1` from the output of `Decompose`.
+ * [FIPS-24] Section 7.4, Algorithm 37.
+ *
+ * This applies the `HighBits` function (Algorithm 37) coefficientwise to
+ * the polynomials in the vectors, per the description at the beginning of
+ * Section 7.4.
+ */
+HighBits : [k]Rq -> [k]R
+HighBits r = r1 where
+    (r1, _) = Decompose r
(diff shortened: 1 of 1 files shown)


mccleeary-galois:
Thank you for breaking this down nicely commit by commit!

PR #208, opened 2024-12-13
Add naive NTT and NTT arithmetic

### LINKED ISSUE #191: ML-DSA: Add NTT-related functions.
We need to define the NTT operations defined in Sections 7.5 and 7.6 in [the ML-DSA spec](https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.204.pdf). The NTT algorithm in the spec looks very similar to the one in ML-KEM, but that one turned out to be very difficult to implement faithfully (see #156). We ended up punting a spec-adherent version to a separate issue (#163). The notes in the IPD version suggest that the NTT algorithms are not identical. On the `reorder` function:
> // Dilithium uses a non-standard ordering of elements in the NTT domain; this function reorders the elements.

We need to make some decisions about how we want to deal with NTT for this algorithm:
- Does this spec's NTT version actually match ML-KEM or is it slightly different? Is the NTT representation different?
- If it's the same, do we want to copy/paste the ML-KEM version or the IPD version?
- Do we want to handle #163 before proceeding here?
- Do we want to try to write a version of NTT now that actually matches the ML-DSA spec?

The ML-DSA spec includes the basic operations `NTT` and `NTTInv`, as well as arithmetic over elements -- and matrices and vectors composed of elements -- in the NTT field `T_q`. Note that the arithmetic operations are not used by name inline; they use overloaded operators ∘ and + to handle all individual / vector / matrix multiplication and individual / vector addition, respectively. 

At least for addition, I think we can use the built-in addition symbol, so we should implement the algorithms as-is and include a property that it's the same as infix addition for both individual elements and vectors.

We could define an infix operator for one of the multiplication functions, but we'd have to choose which one (individual elements, scalar-vector multiplication, or matrix-vector multiplication), and then either use a different infix operator or the function name for the other options. 

- [x] Make decisions about NTT
- [x] Get some version of NTT added and accessible to ML-DSA
- [x] Define named operations on NTT vectors (Algorithms 44 - 48)
- [x] Define addition equality property so we can use the infix operator
- [x] Decide whether to define an infix operator for any of the multiplication functions.

marsella:
Closes #191.

This adds a naive NTT implementation, with citations for its provenance, and some arithmetic over NTT polynomials. I reviewed the spec and added two custom infix operators for the operations that are actually called elsewhere.

### DIFF
--- a/Primitive/Asymmetric/Signature/ML_DSA/Specification.cry
+++ b/Primitive/Asymmetric/Signature/ML_DSA/Specification.cry
@@ -1271,6 +1271,23 @@ AddNTT a b = c where
 property addNTTisPlus a b = AddNTT a b == a + b
 
 /**
+ * Compute the product of two elements in `Tq`.
+ * [FIPS-204] Section 7.6, Algorithm 45.
+ */
+MultiplyNTT : Tq -> Tq -> Tq
+MultiplyNTT a b = c where
+    c = [a@i * b@i | i <- [0..255]]
+
+/**
+ * The definition of multiplication in the spec is the same as the built-in
+ * multiplication (*) operator in Cryptol!
+ * ```repl
+ * :prove multiplyNTTisStar
+ * ```
+ */
+property multiplyNTTisStar a b = MultiplyNTT a b == a * b
+
+/**
  * Compute the sum of two vectors over `Tq`
  * [FIPS-204] Section 7.6, Algorithm 46.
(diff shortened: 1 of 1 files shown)


mccleeary-galois (approved):
Thanks for making this easy to review by commit!

mccleeary-galois (approved):

Issue #192, opened 2024-11-19
ML-DSA: Define top-level internal functions


marsella:
We need to define the `_internal` set of functions for [ML-DSA](https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.204.pdf). 

An important difference from the IPD is that new domain separation was added in Algorithm 6. These three functions don't have direct corollaries in the IPD version, but there is a lot of overlap with the keygen/sign/verify functions in there (labelled algorithm 1, 2, 3).

The sign function in the IPD is separated into several named functions. Generally, I like to contain "subfunctions" in the where clause of the original function, unless we need a top-level feature like numeric constraint guards.

We will want to make sure these are clearly documented that they are not part of the public API

- [x] Define any operations that we don't have written yet, like the infinity norm
- [x] Implement `KeyGen_internal`. Make sure the new domain separation in step 1 is correctly implemented!
- [x] Implement `Sign_internal`. Make sure docs have warnings about no external use, `M'` must be formatted, etc.
- [x] Implement `Verify_internal`. 
- [ ] Define basic properties (running all algorithms on valid input produces valid signatures -- see `dilithiumCorrect` in the IPD).
- [x] Add a KAT (e.g. from [post quantum cryptography](https://github.com/post-quantum-cryptography/KAT/tree/main/MLDSA), if the NIST ones aren't yet available)

PR #211, opened 2024-12-27
ML-DSA: Add sign_internal

marsella:
Addresses part of #192.

This adds `Sign_internal` and several other utility functions that weren't implemented yet. There is one weird thing with endianness happening -- I'll start a discussion inline. 

Chaining this with #209 since it needs the correct NTT implementations to work.

### DIFF
--- a/Primitive/Asymmetric/Signature/ML_DSA/Specification.cry
+++ b/Primitive/Asymmetric/Signature/ML_DSA/Specification.cry
@@ -121,6 +121,30 @@ property modPlusMinusWorks m = inRange && congruent where
     congruent = ((fromZ m) % `α) == (m' % `α)
 
 /**
+ * Compute the infinity norm over a list of elements in `Rq`.
+ * [FIPS-204] Section 2.3, "‖⋅‖∞".
+ */
+infNormRq : {m} (fin m) => [m]Rq -> Integer
+infNormRq w = maxList [ maxList [ abs (modPlusMinus`{q} wij)
+    | wij <- wi]
+    | wi <- w]
+
+/**
+ * Compute the infinity norm over a list of elements in `R`.
+ * [FIPS-204] Section 2.3, "‖⋅‖∞".
+ */
+infNormR : {m} (fin m) => [m]R -> Integer
+infNormR w = maxList [ maxList [ abs wij | wij <- wi] | wi <- w]
+
+private
+    /**
+     * Compute the max of a list of non-negative elements, as in
+     * [FIPS-204] Section 2.3, "‖⋅‖∞".
+     */
+    maxList : {n, a} (fin n, Literal 0 a, Cmp a) => [n]a -> a
+    maxList list = foldl max 0 list
+
+/**
  * Apply `NTT` and `NTTInv` entry-wise over a vector of polynomials.
  * [FIPS-204] Section 2.5.
  *
--- a/Primitive/Asymmetric/Signature/ML_DSA/Specification.cry
+++ b/Primitive/Asymmetric/Signature/ML_DSA/Specification.cry
@@ -177,6 +177,15 @@ H : {l, m} (fin m) => [m][8] -> [l][8]
 H str = SHAKE256::xofBytes`{8 * l} str
 
 /**
+ * Wrapper function around SHAKE256 that takes bit strings as input.
+ * [FIPS-204] Section 3.7.
+ *
+ *
+ */
+HBits : {l, m} (fin m) => [m] -> [l][8]
+HBits str = split (SHAKE256::xof`{8 * l} (str))
+
+/**
  * Wrapper function around SHAKE128, specifying the length `l` in bytes.
  * [FIPS-204] Section 3.7.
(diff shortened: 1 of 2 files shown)


Primitive/Asymmetric/Signature/ML_DSA/Specification.cry
```diff
@@ -307,6 +349,69 @@ KeyGen_internal ξ = (pk, sk) where
...
+
+    // Steps 2 - 5.
+    s1_hat = NTT_Vec (castToRq s1)
+    s2_hat = NTT_Vec (castToRq s2)
+    t0_hat = NTT_Vec (castToRq t0)
+    A_hat = ExpandA ρ
+
+    // Step 6. The endianness does not align with the spec here -- we had to
+    // use `join` instead of `BytesToBits`ing here.
+    μ = HBits`{64} (join tr # M')
```
marsella, line 370: todo: the input to the hash function is supposed to be converted using `BytestoBits` but it only matches the test vectors with `join`. `BytesToBits` maintains byte order of the input but switches the endianness of the bits in each byte. `join` maintains both bit and byte order. 

Other uses of the bit/byte conversion functions from the spec are working as expected. For example, the secret key encode and decode functions depend on the conversion functions. I think they are correct because the secret key output from `KeyGen` matches the KATs.

Thus I'm guessing there is some inconsistency with our expectations around the hash API. In particular, the spec claims that hashing bits should be equivalent to hashing bytes using `BytestoBits` -- but that's not true for us.

Primitive/Asymmetric/Signature/ML_DSA/Tests/ML_DSA_44.cry
```diff
@@ -17,22 +17,26 @@ import Primitive::Asymmetric::Signature::ML_DSA::Instantiations::ML_DSA_44 as ML
...
  *
- * "Deterministic" categorizes how the signing seeds `ρ'` are computed.
+ * "Deterministic" categorizes the per-message randomness parameter `rnd` for
+ * `Sign_internal` -- in this case, it is zero.
  * "Raw" categorizes the set of functions being tested -- in this case, the
  * `{KeyGen, Sign, Verify}_internal` functions. See [PQC-KAT] README.md for
  * more details.
  */
-tryDetKAT xi expected_pk expected_sk = keygenWorks where
+tryDetKAT xi expected_pk expected_sk msg expected_sig = signWorks where
```
marsella, line 26: note: this isn't testing keygen right now. I took it out so my test would run faster during development. I can add that back in now if it seems safer (it does seem safer, now that I think about it...).

Primitive/Asymmetric/Signature/ML_DSA/Tests/ML_DSA_44.cry
```diff
@@ -165,7 +169,7 @@ property det0 = tryDetKAT xi pk sk where
     ])
     msg = 0x6dbbc4375136df3b07f7c70e639e223e
     mlen = 16
-    sm = join [
+    sm = split (join [
```
marsella, line 172: note: the KAT provides the signature appended to the message, but the `sign_internal` function (and also `sign` as far as I can tell) only produces a signature, so I'm not sure why. Anyway, that's what the append above is for and the naming here.

Primitive/Asymmetric/Signature/ML_DSA/Specification.cry
```diff
@@ -307,6 +349,69 @@ KeyGen_internal ξ = (pk, sk) where
     // Step 10.
     sk = skEncode ρ K tr s1 s2 t0
 
+
+/**
+ * Deterministic algorithm to generate a signature for a formatted message `M'`.
+ * [FIPS-204] Section 6.2, Algorithm 7.
+ */
```
marsella, line 356: todo: this likely requires some additional warnings about usage -- e.g. tha the message must be formatted and this shouldn't be called at the top-level except for testing.

Primitive/Asymmetric/Signature/ML_DSA/Tests/ML_DSA_44.cry
marsella: Closed by #212.

Primitive/Asymmetric/Signature/ML_DSA/Tests/ML_DSA_44.cry
marsella: note: I added docs for this in #212.

mccleeary-galois (approved):

Primitive/Asymmetric/Signature/ML_DSA/Specification.cry
marsella: This is partly addressed in #213.

Primitive/Asymmetric/Signature/ML_DSA/Specification.cry
marsella: This was addressed in #212.

Issue #199, opened 2024-12-02
ML-DSA: Define HashML-DSA external functions


marsella:
We need to add the top-level external functions for HashML-DSA. As with the other external functions (#198), these must generate randomness, perform checks, and format inputs. Unlike the other external functions, these allow the user to pass a pre-hashed message, in case the hashing step in signing / verification is prohibitively slow.

This comes with lots of additional caveats, so document those warnings carefully!

Since we can't generate randomness in cryptol, we need to pass it as an `Option` instead (there's an example in ML-KEM).

- [ ] Decide whether to reimplement keygen or somehow include it over from #198.
- [ ] Implement Algorithm 4 `Sign`
- [ ] Implement Algorithm 5 `Verify`
- [ ] Add additional algorithm-level warnings as needed.

PR #213, opened 2025-01-02
ML-DSA: Add external API functions

### LINKED ISSUE #198: ML-DSA: Define external functions
We need to add the top-level external functions for ML-DSA. The external functions are supposed to generate randomness and perform various checks on randomness and other inputs, and format inputs.

Since we can't generate randomness in cryptol, we need to pass it as an `Option` instead (there's an example in ML-KEM).

Structurally, we might want to rearrange / rename things a little bit. All the other functions implemented so far are from section 7, "Auxiliary functions" and section 6 "Internal functions". We want to support this front-end as well as the HashML-DSA front end. It might be good to put the two front-ends into separate modules so the module names look nice and so that each public API can import all the internal & auxiliary functions.

Read the expository content carefully and document any warnings and usage limitations carefully.

- [x] Decide on modules structure and rename things if necessary.
- [x] Implement Algorithm 1 `KeyGen`
- [x] Implement Algorithm 2 `Sign`
- [x] Implement Algorithm 3 `Verify`
- [x] Read Section 3 carefully and add top-level or algorithm-level warnings as needed.
- [x] Add one of the message-formatting KATs from PQC

marsella:
Closes #198 

- Rearranges the module structure a little to better support the public API functions separately from the internal functions, and to eventually make it easier to add the hashed class of functions (#199). 
- Adds the external functions with documentation, including docs about the one deviation with bit ordering.
- Adds some extra KATs


mccleeary-galois (approved):
LGTM, thanks for changing the `:prove` to a `:check`

Issue #179, opened 2024-11-11
Clean up small issues with AES


marsella:
I was reviewing AES for a separate project and identified a few things I'd like to change to bring it closer in line [with the spec](https://doi.org/10.6028/NIST.FIPS.197-upd1), with respect to our [gold standard criteria](https://github.com/GaloisInc/cryptol-specs/wiki/Reviewing-guidelines). Most of these are picky style preferences; none of them affect the functionality -- just readability and obvious spec equivalence.

- [x] Add sbox example from p14 as a property (with a docstring)
- [x] Change sbox to be an actual function instead of an array index
- [x] Move SBox into the Algorithm module 
- [x] Rename `Algorithm` module to `Cipher` to match the spec sections
- [x] Make `AddRoundKey` an explicitly named function instead of an inline `^` and replace all uses
- [x] AES functor: wrap text on the constraint (this is a formatting nit - aim for <80 col)
- [x] Update key expansion functions to use the same naming as the spec.
- [x] Update key expansion to return a single array. Modify call sites (`cipher` and `invCipher`) to decompose internally. Note: the `AES_GCM_SIV` implementation also uses the key expansion types so we can't get rid of them entirely right now; document any types that are only kept for compatibility with that impl.
- [x] Try writing cipher using the `iterate` function and decide if it looks nicer
- [x] Rewrite subword without a list comprehension

marsella:
The last remaining issue about `SBox` is honestly larger than this issue can reasonably contain. I made #238 to capture the actual goals of that task.

PR #218, opened 2025-01-10
AES cleanup fixes pt 2 #179

jn80842:
Two more small fixes for the AES spec from issue #179

### DIFF
--- a/Primitive/Symmetric/Cipher/Block/AES/SBox.cry
+++ b/Primitive/Symmetric/Cipher/Block/AES/SBox.cry
@@ -36,6 +36,7 @@ sbox b = sboxTable @ b
  * Sbox example from p. 14
  * ```repl
  *  :prove sBox53
+ * ```
  */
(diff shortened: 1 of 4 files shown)


Primitive/Symmetric/Cipher/Block/AES/SBox.cry
```diff
@@ -32,6 +32,13 @@ private sboxTable = [ transform (GF28::inverse b) | b <- [0 .. 255] ] where
 sbox : GF28 -> GF28
 sbox b = sboxTable @ b
 
+/**
+ * Sbox example from p. 14
+ * ```repl
+ *  :prove sBox53
```
marsella, line 38: note: There's no closing \`\`\` here! Luckily this feature was designed to work even with it missing, so it doesn't really matter if you fix or not, but FYI :)

marsella (approved):
Looks good!

PR #231, opened 2025-01-28
Speck Improvments

j-christensen1:
Similar to #230 but for Speck. Will be marked as draft while style comments are ported from pull request for Simon.


Primitive/Symmetric/Cipher/Block/Speck/Specification.cry
```diff
@@ -0,0 +1,285 @@
...
+
+    /**
+     * `alpha` is a rotation constant.
+     */
+    type alpha : #
+
+    /**
+     * `beta` is a rotation constant.
+     */
+    type beta : #
```
marsella, line 41: todo: `alpha` and `beta` are dependent on the block size, so these should be constrained appropriately. If they were values I'd say to define them outside the parameter block with a numeric constraint guard. Since they're types, I think they should live here with constraints, like:
```
type alpha : #
type constraint ((n == 16 && alpha == 7) || (alpha == 8))
```

Primitive/Symmetric/Cipher/Block/Speck/Specification.cry
marsella, line 1: todo (style): the definitions in this file don't appear in the same order as in the specification. Please rearrange so they match. I'd expect the order to be, roughly:
1. Circular shift functions `S`
2. The round function and inverse (original formulation)
3. Encryption function
3. Feistel map version of rounds and encryption
5. Key schedule

If properties are explicitly noted in the spec, they should live in that spot. Otherwise you can put them in the first place they make sense (e.g. the equivalence property for the round functions goes after the second round function version is defined).

Primitive/Symmetric/Cipher/Block/Speck/Tests/tests.cry
```diff
@@ -0,0 +1,154 @@
+/**
+ * @copyright Galois, Inc.
+ * @author John Christensen <jchristensen@galois.com>
+ *
+ * Test vectors for Speck.
+ * These test vectors were taken from:
```
marsella, line 6: todo (minor): For consistency you might note "Appendix C" here. Readers will be able to find it but we tend toward aggressive referencing.

Primitive/Symmetric/Cipher/Block/speck.cry
marsella, line 1: note: @sauclovian-g has noted in the past that it's nice to have an explicit commit renaming files so that people poking through the git history can track the changes (e.g. [the history of the ML-KEM spec]( https://github.com/GaloisInc/cryptol-specs/commits/master/Primitive/Asymmetric/KEM/ML_KEM/Specification.cry) -- if you scroll down you'll see the pointer to the old name).

I don't think it's strictly necessary to go back now and change your git history, but please keep in mind in the future.

marsella (commented):
Thank you! Most of these comments are minor! The main important change is to reorder the file to better match the spec.

Primitive/Symmetric/Cipher/Block/Speck/Specification.cry
```diff
@@ -0,0 +1,285 @@
...
+
+    /**
+     * Composition of many functions from `(a -> a)`.
+     */
+    composeMany : {nf, a} (fin n) => [nf](a -> a) -> a -> a
+    composeMany fcns
+        | nf == 0 => \x -> x
+        | nf == 1 => fcns@0
+        | nf == 2 => compose (fcns@0) (fcns@1)
+        | nf > 2  => compose (fcns@0) (composeMany (drop `{1} fcns))
```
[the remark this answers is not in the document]
j-christensen1: I moved these to `Common/utils.cry`. I also made a first pass at documenting the undocumented functions in there, but I made that it's own commit in case we'd like cherry pick that out of this PR.

Primitive/Symmetric/Cipher/Block/Speck/Specification.cry
j-christensen1: I added the weaker constraint `(α >= 7, α <= 8)`, which doesn't resolve the dependency on `n`...are the operators `||` and `&&` type-level? I was having some problems with them.

j-christensen1: Perhaps we define `Sα` and `Sβ` with numeric constraint guards, something like:
```
Sα x | n == 16 => S `{7} x
```
etc...if you'll forgive my syntactic wybbles.

marsella: Oops yeah `||`s aren't allowed in type constraints, my bad. I like the `Sα` definition with constraint guards! That's clearer to me and I think captures the spirit of the constraint better.

PR #232, opened 2025-01-28
Optimize ML-DSA Implementation

### LINKED ISSUE #205: ML-DSA: Optimize implementation
The version of ML-DSA that is currently being developed is very closely adherent to the spec. However, the spec is not written in a way that lends itself to fast Cryptol code. There are two particular types of slow-code that I've noticed:
- direct indexing while operating over arrays; and
- iterating + rebuilding arrays instead of using sequence-level operations (e.g. `join`, `reverse`, etc.)

In general, we don't want to delete the spec-adherent code, because spec adherence is a high-level goal of this repository. We can either make separate functions inline (e.g. `BitsToBytes_fast`) or make a separate module with the fast versions -- this might depend a little on the architecture decision we make in #198. In either case we need to prove equivalence between the spec version and the fast version.

Here are a few notes about things I've seen that could probably be faster:
- `IntegerToBits`: we could use the built-in function `fromInteger`, with a `reverse` call to get the endianness right. Similarly with `BitsToInteger` and probably `IntegerToBytes`
- `BitsToBytes` and `BytesToBits`: these should just be `split` and `join` calls.
- All the `BitPack` functions (Alg 16 - 19) index into an array, but they could iterate directly over it instead.

---

- [x] Audit or benchmark the spec to identify areas for improvement (either based on Cryptol aesthetics or on some actual timing measurements). Put a list in this issue for future reference.
- [x] Decide where to put the "fast" versions.
- [x] Implement the optimized versions.

mariosge:
Closes #205


Primitive/Asymmetric/Signature/ML_DSA/OptimizedSpecification.cry
marsella, line 1: todo: Can you put comments on the functions that are not functionally changed? All the ones from 5949fb4. Just a one-line comment to say, like
```
/**
 * This is unchanged from `Spec`; it's duplicated to pull in the optimized 
 * dependencies in this file.
 */
```

Primitive/Asymmetric/Signature/ML_DSA/OptimizedSpecification.cry
```diff
@@ -0,0 +1,344 @@
...
+skDecodeEquivalence : PrivateKey -> Bit
+property skDecodeEquivalence sk = Spec::skDecode sk == skDecode sk
+
+sigEncode : [λ / 4]Byte -> [ell]R -> [k]R2 -> Signature
+sigEncode c_til z h = σ where
+    // Note that `σ#` indicates the value of `σ` at Step `#`.
+    // Step 1.
+    σ1 = c_til
+    // Step 2 - 4.
+    σ3 = σ1 # join [Spec::BitPack`{γ1 - 1, γ1} (z@i) | i <- [0..ell-1]]
```
marsella, line 221: todo: this should call the fast version, not the spec version!

marsella (changes requested):
Thank you, this looks great! I'd be very interested to see what kind of runtime improvement we get on the top-level functions (`_internal`). It probably won't be 10x because NTT and other operations are slow right now, but that's the big one we're interested in.

I have a couple docs requests and one bug.

Primitive/Asymmetric/Signature/ML_DSA/OptimizedSpecification.cry
```diff
@@ -0,0 +1,344 @@
...
+
+    // Step 8.
+    pk = pkEncode ρ t1
+    // Step 9.
+    tr = H pk
+    // Step 10.
+    sk = skEncode ρ K tr s1 s2 t0
+
+KeyGen_internalEquivalence : [32]Byte -> Bit
+property KeyGen_internalEquivalence ξ = Spec::KeyGen_internal ξ == KeyGen_internal ξ
```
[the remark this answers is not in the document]
mariosge: I've addressed this! I believe I used the same values for the types parameters as the ones from the properties in the specs.

Primitive/Asymmetric/Signature/ML_DSA/OptimizedSpecification.cry
mariosge: Great catch! I've addressed this as well!

marsella (approved):
Looks good! Thanks for updating.

PR #233, opened 2025-01-30
Replace `mp_mod_inv` with `recip`

### LINKED ISSUE #100: Add prime constraint to inverse function
The `mp_mod_inv` method in `Common/utils.cry` is used in several places, but its type signature is not correct; it's only a valid function when the modulus is prime. Cryptol has evolved since that function was first written in 2018, so now I see two options to make it correct:
1. Add a `prime p` constraint to the function type constraint
2. Replace all uses with the `recip` and/or `/.` cryptol-native operators on `Z` types.

I did a quick review of all the modules that use this and it looks like they do all use it correctly (e.g. with a prime modulus). If we choose option 1, there will be some cascading work to add the `prime` constraint up the call chain, but that seems worth the precision to me.

That said, if the built-in operations do what I think they do (compute the multiplicative inverse and do modulus-respecting division), we should use those. We might consider improving documentation on those as well, to be explicit about any additional constraints they may have (e.g. does the modulus have to be prime for those operations? I don't know).

- [ ] Pick an option to make the inverse function correct
- [ ] Propagate it throughout the codebase
- [ ] Update any missing documentation / add useful clarifications where necessary

RyanGlScott:
Closes #100.

`mp_mod_inv` has an incorrect type signature, as it is only valid when the modulus is prime, but the type does not reflect this. Moreover, there is now a correct version of this in the Cryptol prelude in the form of `recip : {p} (prime p) => Z p -> Z p`, so there is really no need to define `mp_mod_inv` separately anymore. Let's just remove `mp_mod_inv`.

The only real call site for `mp_mod_inv` was in `Common::EC::PrimeField::PFEC`, but it is easily swapped out for `recip`.

### DIFF
--- a/Common/utils.cry
+++ b/Common/utils.cry
@@ -66,39 +66,3 @@ half x = fromInteger (if isEven xint then xint/2 else ((xint + `p) / 2))
 
 half_correct : {p} (fin p, p >= 3, isOdd p) => Z p -> Bit
 property half_correct x = half x + half x == x
-
-/**
- * Binary algorithm for inversion in Z n, for a prime `n`.
- *
- * [HMV] Algorithm 2.22. [MATH-2008] recommends this algorithm because all the
- * divisions are by 2, so they could be implemented as right shifts. In this
- * implementation, we don't take advantage of that optimization.
- *
- * [HMV]: D. Hankerson, A. Menezes, S. Vanstone, Guide to Elliptic Curve
- * Cryptography. Springer-Verlag, New York, 2004.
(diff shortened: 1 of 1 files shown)


marsella (approved):
This looks great, thank you! I peeked into the history and this used to be used in some of the old ECDSA and EC instances that were deleted in #113 and #174. Happy to have it gone.

weaversa:
When I wrote that function 7 years ago (!) we didn't have a `prime` type constraint. I'm glad to see it is no longer needed.

PR #235, opened 2025-01-30
`PFEC`: Only compute `to_affine`'s `lambda` when necessary

### LINKED ISSUE #234: `PFEC`: Refactor `to_affine` to only compute reciprocal only when necessary
The current definition of `to_affine` in `Common::EC::PrimeField::PFEC` is:

https://github.com/GaloisInc/cryptol-specs/blob/5cda1c7c9fc7018edcbe5c5e315a5f6f8b30bb96/Common/EC/PrimeField/PFEC.cry#L356-L366

Note that `lambda` (which is bound in a `where` clause) is only used in the `else` branch of the `if`-expression. Critically, `lambda` should only be evaluated if `p.z != 0`, as `recip p.z` will panic if `p.z == 0`. With Cryptol's lazy evaluation, this is not an issue, but it does become an issue when the Cryptol code is compiled to call-by-value languages (e.g., C or Rust), where the resulting code looks more like:

```rs
let lambda = mp_mod_inv(p.z);
if p.z == 0 {
  Point::Infinity
} else {
  Point::Affine(exp(lambda, 2) * p.x, exp(lambda, 3) * p.y)
}
```

Thankfully, it is very straightforward to avoid this problem by making a small tweak to the code:

```diff
diff --git a/Common/EC/PrimeField/PFEC.cry b/Common/EC/PrimeField/PFEC.cry
index 4dca378..1d9586b 100644
--- a/Common/EC/PrimeField/PFEC.cry
+++ b/Common/EC/PrimeField/PFEC.cry
@@ -361,9 +361,9 @@ private
     to_affine p =
         // The point at infinity does not have an affine representation.
         if p.z == 0 then Infinity
-        else Affine (lambda ^^2 * p.x) (lambda ^^3 * p.y)
-        where
-            lambda = recip p.z
+        else (Affine (lambda ^^2 * p.x) (lambda ^^3 * p.y)
+               where
+                 lambda = recip p.z)
 
 
     // The twin affine function depends on the set of four points: two
```

This is such a minor change that I am inclined to simply adopt it upstream.

RyanGlScott:
Fixes #234.

### DIFF
--- a/Common/EC/PrimeField/PFEC.cry
+++ b/Common/EC/PrimeField/PFEC.cry
@@ -361,9 +361,17 @@ private
     to_affine p =
         // The point at infinity does not have an affine representation.
         if p.z == 0 then Infinity
-        else Affine (lambda ^^2 * p.x) (lambda ^^3 * p.y)
-        where
-            lambda = recip p.z
+        else (Affine (lambda ^^2 * p.x) (lambda ^^3 * p.y)
+               // We use parentheses to limit the scope of this `where` clause
+               // to the `else` branch of the `if`-expression. Doing so is not
+               // strictly necessary in Cryptol, as Cryptol's lazy evaluation
+               // will only evaluate `lambda` if the `else` branch is evaluated.
+               // This is solely done for the benefit of compiling to
+               // call-by-value languages (e.g., C or Rust), as evaluating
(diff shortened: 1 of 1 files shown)


marsella (approved):
Thanks for the added documentation!

PR #237, opened 2025-02-04
AES cleanup fixes pt 3 issue 179

jn80842:
Remaining small fixes from issue [179](https://github.com/GaloisInc/cryptol-specs/issues/179)


marsella (commented):
This looks great! One return type I think needs to change but otherwise excellent.

marsella (approved):
Thanks for fixing, looks good!

PR #244, opened 2025-02-12
Remove use of Integer in md5

glguy:
Integers are hard to support in some compiler targets and this particular parameter only needs to work with up to 5-bit numbers, as far as I can tell

### DIFF
--- a/Primitive/Keyless/Hash/MD5.md
+++ b/Primitive/Keyless/Hash/MD5.md
@@ -194,7 +194,7 @@ T =
 Do the following:
 
 ```cryptol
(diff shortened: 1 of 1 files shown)


marsella:
This seems reasonable to me but I'm confused about why it breaks the unrelated SPHINCS+ spec.

marsella (commented):
Please add a comment with the copyright info to the top of this file. I will approve once that check passes, since the other one is unrelated and caused by a change in nightly cryptol, not this PR.
```
<!---
 - @copyright Galois, Inc.
 - @author Alannah Carr
 --->
 ```

marsella (approved):
Thanks!

PR #246, opened 2025-02-18
Add cryptol project for Suite B

### LINKED ISSUE #241: Add Cryptol-Projects to Suite B algorithms
With the completion of https://github.com/GaloisInc/cryptol/issues/1334, cryptol projects are now available! We should add them to this repo. The Suite B algorithms are self-contained and also self-referential (e.g. ECDSA uses SHA2), so they are a good demonstration of the projects interface. [Documentation on projects is in the reference manual](https://galoisinc.github.io/cryptol/master/Project.html).

- [x] Add `.toml` files for the specifications in suite B
    - [x] AES
    - [x] SHA2
    - [x] ECDH
    - [x] ECDSA
- [x] Update documentation (maybe at the spec level?) to note the use of projects
- [ ] If you generate any opinions, add style notes for writing `.toml` files to #5.

marsella:
Closes #241.

Adds a `.toml` file with [the Cryptol project definition](https://galoisinc.github.io/cryptol/master/Project.html) for all the algorithms in Suite B, plus a few bonus parameter sets.

In terms of best practices, I did not push the cache here. I figured the project file should exist for reference, but any actual caching should be maintained by CI.

I was thinking about whether it would be nice longer-term to include separate project specifications for different algorithm sets (for example, a `suiteb.cryproject.toml` and a `cnsa2.cryproject.toml`, and maybe another for post-quantum algorithms). I think this would require extending cryptol projects to support multiple caches, named caches, and if we really want to optimize, some kind of shared cache (since e.g. AES is in both Suite B and CNSA 2.0).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,6 @@
+## Generated files from Cryptol projects:
+loadcache.toml
+
 ## Core latex/pdflatex auxiliary files:
 *.aux
 *.lof
--- a/Primitive/Symmetric/Cipher/Block/AES/ExpandKey.cry
+++ b/Primitive/Symmetric/Cipher/Block/AES/ExpandKey.cry
@@ -65,8 +65,8 @@ Rcon j = constants @ (j - 1) where
  * :prove RconIsExponentiation
  * ```
  */
-RconIsExponentation : [8] -> Bit
-property RconIsExponentation j = (1 <= j) && (j <= 10) ==>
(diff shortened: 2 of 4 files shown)


mccleeary-galois:
LGTM, makes sense to put the cache in CI.

mccleeary-galois (approved):

Issue #238, opened 2025-02-05
Update structure for AES


marsella:
The structure of AES is currently kind of scattered. The instantiations live in the `Cipher/Block/` directory, as does the top-level API, but the actual content is spread across several modules in the `Cipher/Block/AES/` directory. I think this should be consolidated to match the structure of the other gold-standard specs in our repo.
```
Primitive/Symmetric/Cipher/Block/AES/
|-- Specification.cry
|-- Instantiations/
|-- Tests/
```

Here are the steps I see at a first look. If other things come up, please note them in this issue. 

- [x] Consolidate the whole specification into one file. This will contain everything in `AES_specification` and the `AES/` directory. The only parameter should be the key size; other parameters (`Nk`, `Nr`, `Mode`) can be computed from that value. Order should be preserved according to the NIST spec.
- [x] Keep the public API as-is for now, but try to simplify. In particular:
    - [x] Note on `encrypt` and `decrypt` that these are common names but the spec uses different names, add citation to Equation 5.1. (Could consider exporting equivalent `AES` and `AESInv` functions, but the spec is not super clear on these names so fine to skip).
    - [x] Remove the `decryptWithSchedule` function
    - [x] Add docs on `encryptWithSchedule` and `keyExpansion` that they're public for legacy reasons and shouldn't be used in general.
- [x] Update READMEs and fix links (both top-level and cipher-level READMEs).
- [x] Fix dependencies (e.g. in everything that uses concrete AES instantiations)
- [x] Decide where to put instantiations and tests for modes of operation. I vote either in `Cipher/AES/Instantiations` (so that would have both AES128 and AES128_GCM) or in `Cipher/Modes/Instantiations/` (which might have multiple ciphers in it).

jn80842:
Resolved with PR #250.

PR #249, opened 2025-02-28
Update structure for AES

jn80842:
Address issue #238 by consolidating the AES files to match the structure of the other gold-standard specs in our repo.


jn80842:
See PR #250 instead.

PR #250, opened 2025-03-03
reorganize AES file structure to match other gold-standard specs

jn80842:
Consolidates the AES structure to match the other gold-standard specs; addresses issue #249


marsella:
Thank you for taking this on!

It's hard to tell from this diff whether there were any content changes made in the consolidation step 059d487. Git also lost the thread of names here. Could you make a couple of changes in the commit history:
1. Add an initial commit that only renames the biggest file (probably `Cipher` to `Specification`).
2. Add an individual commit for each file to paste it into `Specification`. Note in the commit message that there are no functional changes (unless names need to change because of overlap).
4. Add a commit to do any renaming API and naming changes for clarity and consistency with the NIST doc.

Sorry, I know it's a hassle! I appreciate it though.

jn80842:
Okay, I restructured the commit structure as requested.

Also, the copyright check is failing on a markdown file -- how do we add copyright info to those?

marsella:
Thank you, looks much better!

You can add comments to markdown using the HTML notation `<!-- -->`; there's [an example in MD5](https://raw.githubusercontent.com/GaloisInc/cryptol-specs/b19c487b73cc5f5a7e8790fa7e5640f18337addc/Primitive/Keyless/Hash/MD5.md).

I usually poke back [in the git history](https://github.com/GaloisInc/cryptol-specs/commits/b19c487b73cc5f5a7e8790fa7e5640f18337addc/Primitive/Symmetric/Cipher/Authenticated/SIV_rfc5297.md) to find everyone who's worked on a file. At this point we're pretty generous with authorship as long as someone touched the file; I occasionally mark `@editor` if there's only been a small change but it's a soft distinction.

Issue #104, opened 2024-08-01
Add more valid ECDSA test vectors


marsella:
There are a bunch of test vectors for ECDSA in [RFC 6979](https://datatracker.ietf.org/doc/html/rfc6979). Some of these are for invalid curve-hash combinations. Decide which of these we'd like to include in the repo, then instantiate the needed versions of ECDSA and implement the test vectors.

This might be blocked on #98, in order to be able to switch out different hash functions easily. The current version fixes it to SHA-256, but that only works with two of our curves (P-224 and P-256).

- [x] Decide which hash-curve combos to instantiate
- [x] Add all the 6979 test vectors
- [x] Do a time-limited search for other test vector sources

marsella:
Here's a [permalink](https://github.com/GaloisInc/cryptol-specs/blob/5dae3b3b58f6930a5be5df5f6eb88f10bce596be/Primitive/Asymmetric/Signature/ECDSA/ECDSA_sign_tests.cry) to the ECDSA sign tests from the previous version of the implementation.

marsella:
#32 also references some [ECDSA test vectors](https://github.com/GaloisInc/cryptol/tree/master/tests/suiteb) from NIST that are already in Cryptol syntax. Add these too, if appropriate!

PR #254, opened 2025-03-10
Instantiate ECDSA-P521-SHA512

glguy:
@marsella This PR needs help. The specification module for ECDSA doesn't like instantiating a 521-bit curve together with a 512-bit digest, but this seems to be more or less the two that get paired together. How do you think we should proceed on this one?

We'd probably also want to instantiate ECDSA-P521-SHA3-512 as well, but given that the test vectors were less obvious for that one and there's this type constraint question I didn't want to get too far ahead.


marsella:
Right, this is a bit of a clash between the official line in the spec and the reality of the parameter choices. The spec says:
> It is recommended that the security strength associated with the bit length of n and the security
strength of the hash function be the same unless an agreement has been made between
participating entities to use a stronger hash function. A hash function that provides a lower
security strength than is associated with the bit length of n **shall not** be used. ([FIPS 186-5, Section 6.1.1](https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.186-5.pdf))

This is because they don't want people to accidentally dilute a nice strong elliptic curve with a weak hash function (and not realize it). However, the P521 curve is awkward because its security strength is greater than the usual cutoff of 512 bits, so using a slightly shorter hash function in this case is usually fine. 

We can get around this by using the ECDSA `UnconstrainedSpec` instead of `Specification`, which drops the requirement that the hash strength is no less than the curve strength.

Primitive/Asymmetric/Signature/ECDSA/Instantiations/ECDSA_P521_SHA3_512.cry
```diff
@@ -0,0 +1,15 @@
+/*
+ * Instantiate ECDSA with curve P-521 (also known as secp521p1) and SHA3-512.
+ *
+ * Note that this uses `UnconstrainedSpec` because the security strength of the hash
+ * function is less than that of the elliptic curve! The overall security strength
+ * of this ECDSA instantiation is approximately 256 bits.
+ *
+ * @copyright Galois, Inc.
+ * @author Marcella Hastings <marcella@@galois.com>
```
glguy, line 9:
```suggestion
 * @author Marcella Hastings <marcella@galois.com>
```

glguy:
I can't *review* my own pull request, even though you've now done most of the work on it. I would have clicked **Approve**.

marsella (approved):
Nice, looks good to me too. Linking #104 since that has been used historically to add more ECDSA instantiations.

Issue #129, opened 2024-08-29
Add more NIST test vectors for AES


marsella:
#32 has a pointer to [some NIST test vectors for AES](https://github.com/GaloisInc/cryptol/tree/master/tests/suiteb) that have been expertly munged into Cryptol syntax.

- [ ] Double check that these are different from the test vectors we currently have for AES
- [ ] Add them to the repo if so
- [ ] Add appropriate citations
- [ ] Add docstrings to check the new vectors

j-christensen1:
See #264. There are still `mct` (Monte Carlo tests) to consider.

PR #264, opened 2025-03-24
AES: Add CAVP ECB test vector examples.

j-christensen1:
Relates to #129


marsella (approved):
This looks good! I ran `:check-docstrings` locally on all three of these and they passed!

PR #265, opened 2025-03-25
SHA3: Attempt at Bugfix

j-christensen1:
Appeal to `mod` as a homomorphism from `Integer` to Z 256 ~~ `[8]`, since `ir` is computed with addition and multiplication..

### DIFF
--- a/Primitive/Keyless/Hash/SHA3/Specification.cry
+++ b/Primitive/Keyless/Hash/SHA3/Specification.cry
@@ -282,7 +282,7 @@ private
      *
      * [FIPS-202] Section 3.2.5, Algorithm 5.
      */
-    rc : Integer -> Bit
+    rc : [8] -> Bit
     rc t = if (t % 255) == 0 then 1 // Step 1.
          else Rs @ (t % 255) @0 // Step 4.
          where
@@ -313,7 +313,7 @@ private
      *
      * Equivalent to [FIPS-202] Section 3.2.5, Algorithm 5.
      */
-    rc_hardcoded : Integer -> Bit
+    rc_hardcoded : [8] -> Bit
     rc_hardcoded t = constants @ (t % 255) where
         constants = join [
             0x80b1e87f90a7d57062b32fde6ee54a25,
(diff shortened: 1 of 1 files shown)


j-christensen1:
I believe my appeal to homomorphism is sufficient to resolve the bug. Please review.

Primitive/Keyless/Hash/SHA3/Specification.cry
```diff
@@ -344,7 +344,12 @@ private
         // Step 3. 0 is the identity for XOR, so we iteratively set the
         // `2^j-1`th element by XORing the previously-set bits of RC with the
         // correct value at the desired index.
-        RCs = [RC0] # [RC' ^ (zext`{w} [rc_hardcoded (toInteger j + 7 * ir)] << (2^^j - 1))
+        // `mod` is a homomorphism between `Integer` and `[8]`.
+        // This means that `mod` "commutes" with addition and multiplication of `Integer`.
+        // Because of this, it is safe, from the perspective of correctness,
+        // to consider `ir` as a `[8]`.
+        ir' = fromInteger ir :[8]
+        RCs = [RC0] # [RC' ^ (zext`{w} [rc_hardcoded (j + 7 * ir')] << (2^^j - 1))
```
marsella, line 352: My concern about this solution is that an 8-bit number is the same as taking an integer mod 256, not 255. So in this line, the addition and multiplication could theoretically mess us up if we end up with a number >= 255.

In practice with the standardized parameters, I think this case never happens. At least, someone else proposed this fix and I think it didn't break any tests (presumably you also ran the KATs).

However, I'd rather try to come up with a version that won't break for any parameter sets, even if there is overflow here.

j-christensen1: Ah, I understand. I'll give it another go soon.

j-christensen1: I changed the site at which conversion to `[8]` happens, and wrote some properties justifying the change.

marsella (approved):
Thank you, this looks great to me!

Issue #128, opened 2024-08-29
Add check-docstrings to CI


marsella:
I'm starting to make changes to interfaces that cascade down to multiple files and I'm less confident that I'm checking all the the properties in all the dependencies still work. I'd like to add a CI run that addresses this concern without being outrageously slow.

My ideal version would
1. Identify all the Cryptol [files that changed](https://github.com/GaloisInc/cryptol-specs/blob/master/.github/workflows/copyright_check.yml#L24)
2. Recursively identify all files that depend on each changed file
3. Load and run `:check-docstrings` on each file in the set (1 + 2). In parallel, ideally.

I'm not sure what the best way to do 2 is. We have pretty well specified module path names so it might be as straightforward as grepping the whole repo for each file name. There are also some built-in tools that identify dependencies (`:module-deps`) for a given file, but we'd have to do that "in reverse" (that is, we want the files that depend on `X`, not the dependencies of `X`). There might be another way to use haskell / cryptol internals to get the dependency tree.

Related: #56.

- [ ] Choose an approach for this bit of CI
- [ ] Write all the scripts + yml files

mccleeary-galois:
We also need to ensure that we are targeting the solvers that we ship with https://github.com/GaloisInc/what4-solvers/releases/tag/snapshot-20240212 for now. More discussion across ecosystem as a whole needs to happen for solver versions in general.

Isweet:
The concerns expressed by @marsella come up pretty often, and I wonder if it warrants revisiting and prioritizing work on a Cryptol build system and package manager. See, e.g., https://github.com/GaloisInc/cryptol/issues/1334.

Isweet:
> The concerns expressed by @marsella come up pretty often, and I wonder if it warrants revisiting and prioritizing work on a Cryptol build system and package manager. See, e.g., [GaloisInc/cryptol#1334](https://github.com/GaloisInc/cryptol/issues/1334).

It looks like @yav and @glguy are actively working on this, so maybe you can use it here when they merge the initial prototype. See: https://github.com/GaloisInc/cryptol/pull/1526.

j-christensen1:
A first draft of this was completed in #268. There is still room for improvement (Python remote API vs bash?), and there are some issues to resolve (see #272).

PR #268, opened 2025-04-01
Add `check-docstrings` to CI

j-christensen1:
Relates to #128

### DIFF
--- a/.github/workflows/check_docstrings.yml
+++ b/.github/workflows/check_docstrings.yml
@@ -73,7 +73,7 @@ jobs:
           git config --global --add safe.directory '*'
           echo "Check which caches exist, for sanity."
           gh cache list
-          # bash scripts/check_docstrings.sh .
+          bash scripts/check_docstrings.sh .
           # if the cache doesn't exist for this branch, no sweat!
           # we will just upload the one we create.
           set +e
--- a/.github/workflows/check_docstrings.yml
+++ b/.github/workflows/check_docstrings.yml
@@ -76,10 +76,8 @@ jobs:
           bash scripts/check_docstrings.sh .
           # if the cache doesn't exist for this branch, no sweat!
           # we will just upload the one we create.
-          set +e
           echo "Deleting cache key: ${{ env.CRYPTOL_CACHE_KEY }}"
-          gh cache delete ${{ env.CRYPTOL_CACHE_KEY }}
-          set -e
+          gh cache delete ${{ env.CRYPTOL_CACHE_KEY }} || true
       # Cache the .cryproject/* files.
       # Cache is scoped to branch, commit, and cache version.
       # Note that GITHUB_HEAD_REF only available on pull requests.
--- a/.github/workflows/check_docstrings.yml
+++ b/.github/workflows/check_docstrings.yml
@@ -16,6 +16,10 @@ on:
 jobs:
   check-docstrings:
     runs-on: ubuntu-latest
+    # Permissions are required to delete the cache key.
+    permissions:
+      actions: write
+      contents: read
     env:
       # This is the relative path to the Cryptol project directory.
       CRYPTOL_PROJECT: .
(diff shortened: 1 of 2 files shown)


j-christensen1:
A first take at this PR. This implementation leverages `cryproject.toml` files and associated features of Cryptol 3.3.0.

.github/workflows/check_docstrings.yml
```diff
@@ -0,0 +1,69 @@
...
+        uses: actions/cache/restore@v4
+        if: steps.load-cache.outputs.cache-hit != 'true'
+        with:
+          path: ${{ env.CRYPTOL_CACHE }}
+          key: ${{ env.CRYPTOL_DEFAULT_KEY }}
+      # Run the check-docstrings script, regardless of cache load.
+      - name: "Run the check-docstrings script."
+        run: |
+          # Right now there is no cache. Add this later.
+          bash scripts/check_docstrings.sh .
```
marsella, line 59: Flagging this -- is this still true? It looks like the cache is getting restored but not used, but I"m not sure if that's because it isn't correct or if it just isn't sitting in the right place.

.github/workflows/check_docstrings.yml
marsella, line 1: It looks like the cache isn't getting saved (there's [a warning](https://github.com/GaloisInc/cryptol-specs/actions/runs/14202553008)). Might be because there was a failed run sometime during your testing and setup -- [possibly relevant](https://github.com/shivammathur/cache-extensions/issues/22), maybe you have to update the cache key or manually delete it from somewhere?

marsella (commented):
Looks like the caching isn't working yet -- at least, this cache didn't get saved.

.github/workflows/check_docstrings.yml
j-christensen1: Yikes. You're definitely right. It seems [another relevant issue is here](https://github.com/actions/toolkit/issues/505). I'll take a look at some of the workarounds for updating the key.

j-christensen1: Manual reading failure on my part: [see here](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/caching-dependencies-to-speed-up-workflows#using-the-cache-action).

j-christensen1: I'm not entirely pleased with what I came up with to deal with the above, but I now use the [github CLI tool `gh`](https://github.com/cli/cli/tree/trunk) to delete the cache key after a successful run.

.github/workflows/check_docstrings.yml
j-christensen1: It needed quite a bit of love, but on the latest test, the cache is now restored and used. The crux of the issue, it seems, was confusion on my part of how GitHub runners handle `env` and such.

.github/workflows/check_docstrings.yml
```diff
@@ -0,0 +1,92 @@
...
+          git config --global --add safe.directory '*'
+          echo "Check which caches exist, for sanity."
+          gh cache list
+          bash scripts/check_docstrings.sh .
+          # if the cache doesn't exist for this branch, no sweat!
+          # we will just upload the one we create.
+          set +e
+          echo "Deleting cache key: ${{ env.CRYPTOL_CACHE_KEY }}"
+          gh cache delete ${{ env.CRYPTOL_CACHE_KEY }}
+          set -e
```
marsella, line 82: To confirm, this will not bork if the cache didn't exist? It'll fail gracefully? You might add `|| true` at the end to ensure that a failed delete due to a non-existent cache doesn't break the job.

.github/workflows/check_docstrings.yml
marsella: Ah, good catch. I think that's a reasonable approach. I guess the default GH cache action is not designed for something quite this fine-grained and updated, but this seems like a fine workaround.

.github/workflows/check_docstrings.yml
marsella: Oh I just didn't know what `set -e` did. Thank you!

marsella (approved):
Looks good, thanks for adding!

PR #269, opened 2025-04-01
Fix invalid links in README.md

amaloz:
Two of the links, related to AES256 and AES128-GCM, pointed to non-existent cryptol files. This PR fixes that.


marsella (approved):
Thank you for fixing!

PR #278, opened 2025-04-09
Update block cipher readme

marsella:
Moves it to the correct directory and updates the simon and speck indicators.


j-christensen1 (approved):
LGTM!

PR #279, opened 2025-04-10
Fix bugs in EC properties

### LINKED ISSUE #276: Look into EC curve operation special cases
We identified several counterexamples to PFEC properties:
```
Common::EC::PrimeField::PFEC> normalizeWorksOnNormalPoints 0 0 0 0
False
Common::EC::PrimeField::PFEC> twin_mult_works 115792089210356248762697446949407573530086143415290314195533631308867097853950 115792089210356248762697446949407573529996955224135760342422259061068512044368 0 0
False
```

We need to determine where the problem lies:
- is there an incorrect assumption in the properties that allows zero points when it should not?
- is there a special case in the `twin_normalize` function that we're not accounting for?
- Is there a bug in the implementation?

Tasks
- [x] Determine if we have a mistaken assumption or if there's something wrong with the implementations
- [x] Fix appropriately
- [x] Update any docs to make sure the new constraints are captured
- [x] Add a special-case property to capture the correct behavior, if appropriate.

marsella:
This fixes two errors we uncovered in the properties on the prime field elliptic curve implementation. Note that neither of these were problems with the functions themselves -- just mistaken assumptions / function calls in the tests.

I fixed the two bugs and added additional `:check`s to the docstrings to handle the specific cases that were failing.

Closes #276.

### DIFF
--- a/Common/EC/PrimeField/PFEC.cry
+++ b/Common/EC/PrimeField/PFEC.cry
@@ -760,13 +760,19 @@ private
      * Checks that twin multiplication works as expected. Note that a given
      * point can have multiple projective representations, so we have to
      * check equality in the affine representation.
+     *
+     * These tests check the general case, the case where the two points `S`
+     * and `T` are the same, and the case where one of the points is the point
+     * at infinity.
      * ```repl
      * :check twin_mult_works
+     * :check (\d0 d1 k -> twin_mult_works d0 k d1 k)
+     * :check (\d0 d1 k -> twin_mult_works d0 k d1 0)
      * ```
      */
     property twin_mult_works d0 k0 d1 k1 = affineEq R_plain R_twin
         where
             S = to_projective (validPointFromK k0)
             T = to_projective (validPointFromK k1)
-            R_plain = to_affine (ec_add (ec_mult d0 S) (ec_mult d1 T))
+            R_plain = to_affine (ec_full_add (ec_mult d0 S) (ec_mult d1 T))
             R_twin = to_affine (ec_twin_mult d0 S d1 T)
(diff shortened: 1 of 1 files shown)


marsella:
I rebased onto #281 to get the CI running again.

Common/EC/PrimeField/PFEC.cry
```diff
@@ -443,12 +443,21 @@ private
...
-            all_are_normal = (out.A.z == 1) && (out.B.z == 1)
-                && (out.C.z == 1) && (out.D.z == 1)
-            all_match = (out.A == p1) && (out.B == p2)
-                && (out.C == p3) && (out.D == p4)
+            // If any of the multipliers is 0, the resulting point
+            // will be the point at infinity.
+            // In this case, `twin_normalize` will do nothing...
+            any_are_infinity = any (\k -> k == 0) [k1, k2, k3, k4]
 
+            // ...and the point at infinty always has a z-coordinate of 0.
```
amaloz, line 451: Typo here: `infinty` -> `infinity`.

Common/EC/PrimeField/PFEC.cry
```diff
@@ -751,13 +760,19 @@ private
...
      * point can have multiple projective representations, so we have to
      * check equality in the affine representation.
+     *
+     * These tests check the general case, the case where the two points `S`
+     * and `T` are the same, and the case where one of the points is the point
+     * at infinity.
      * ```repl
      * :check twin_mult_works
+     * :check (\d0 d1 k -> twin_mult_works d0 k d1 k)
+     * :check (\d0 d1 k -> twin_mult_works d0 k d1 0)
```
amaloz, line 770: Is it worth adding `:check (\d0 d1 k -> twin_mult_works d0 0 d1 k)`?

amaloz (approved):

Common/EC/PrimeField/PFEC.cry
marsella: Yeah, these are supposed to be symmetric. Added.

PR #281, opened 2025-04-10
ci: fix install of gh command #280

### LINKED ISSUE #280: Install later version of gh command in actions
The cryptol nightly image is now running on Ubuntu 22.04 (previously it was on 24.04 -- see https://github.com/GaloisInc/cryptol/issues/1840). The default `gh` command installed with `apt` on that distro is `v2.4.0+dfsg1`, which doesn't include the `cache` command, which we need to run the check-docstrings CI job. Instead, we should install the `gh` binary directly [from the latest release](https://github.com/cli/cli/releases/tag/v2.69.0) so we're guaranteed to get the command we want.

- [ ] Update [installation of `gh` command](https://github.com/GaloisInc/cryptol-specs/actions/runs/14388628165/workflow?pr=279#L47)

marsella:
Closes #280 -- installs a recent version of gh instead of a potentially-outdated one from `apt`.

### DIFF
--- a/.github/workflows/check_docstrings.yml
+++ b/.github/workflows/check_docstrings.yml
@@ -43,8 +43,10 @@ jobs:
       # We require git to run the github CLI tool.
       - name: Installing dependencies..
         run: |
-          apt update
-          apt install -y git gh
+          apt-get update && apt-get install -y git wget
+          wget https://github.com/cli/cli/releases/download/v2.69.0/gh_2.69.0_linux_amd64.tar.gz
+          tar xzvf gh_2.69.0_linux_amd64.tar.gz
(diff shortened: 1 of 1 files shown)


j-christensen1 (approved):
Looks good to me! When the nightly cryptol image OS version stabilizes, maybe we'll take another look. I know nothing about the cadence that `gh` updates.

marsella:
Yeah, that's fine by me. Arguably this is a better way to do it since we're pinning to a version that we know supports what we need, rather than depending on a relatively unstable image (even outside of the ubuntu changes, `cryptol` is updated on a somewhat regular basis).

PR #286, opened 2025-04-14
`PFEC`: only call `recip` in `twin_normalize` when necessary

RyanGlScott:
Addresses the comment in https://github.com/GaloisInc/cryptol-specs/issues/234#issuecomment-2802717254


marsella (approved):
I reviewed the `PFEC` spec and it looks like these are the only two instances of calling `recip`, but there are other places where the `where` clause is only relevant for the `else` but is attached to the whole conditional, like `ec_double` and `ec_add`. I was thinking about proactively parenthesizing those, but since that's likely just an optimization (until proven otherwise) I think I'll leave it alone.

PR #290, opened 2025-04-16
XMSS helper functions

### LINKED ISSUE #175: XMSS: Implement components
The [XMSS scheme](https://datatracker.ietf.org/doc/html/rfc8391#section-4.1) has several component algorithms and definitions. We'll define these first, before getting to the actual signature scheme.

- [x] Define the anonymous interface, including parameters (`h`, `n`, and `w`) and hash functions (`H` and `H_msg`) - Section 4.1.1 and 4.1.2.
- [ ] Define randomized tree hashing function `RAND_HASH` - Section 4.1.4
- [ ] Define L-tree address and implementation - Section 2.5 and Section 4.1.5
- [ ] Define tree hash routine - Section 4.1.6

amaloz:
Closes #175.

### DIFF
--- a/Primitive/Asymmetric/Signature/WOTS/Interface.cry
+++ b/Primitive/Asymmetric/Signature/WOTS/Interface.cry
@@ -86,6 +86,24 @@ type constraint (width len <= 32)
 type Byte = [8]
 
 /**
+ * Convert a non-negative integer to a binary representation in big-endian
+ * byte order.
+ * [RFC-8391] Section 2.4.
+ *
+ * Note: the spec is not explicit about what to do when `m < 8y` -- e.g. when
+ * converting to `y` bytes means truncating the integer representation,
+ * rather than padding it with 0s. We chose this behavior based on the
+ * reference implementation that accompanies the spec.
+ * @see https://github.com/XMSS/xmss-reference/blob/master/utils.c#L6
+ * [RFC-8391] Section 7.
+ *
+ * Essentially, we pad with zeros if needed (using `zext`), then `drop` any
+ * unnecessary padding (and potentially also drop some bits of the original
+ * integer).
+ */
+toByte : {y, m} (fin y, fin m) => [m] -> [y]Byte
+
+/**
  * Address scheme for randomizing hash function calls in the OTS scheme.
  * [RFC-8391] Section 2.5.
  */
--- a/Primitive/Asymmetric/Signature/WOTS/Interface.cry
+++ b/Primitive/Asymmetric/Signature/WOTS/Interface.cry
@@ -163,8 +163,9 @@ type Signature = [len][n]Byte
 /**
  * Generate a WOTS+ public key from a private key.
  * [RFC-8391] Section 3.1.4, Algorithm 4 (called `WOTS_genPK` in the spec).
+ * [Errata-8382] @see https://www.rfc-editor.org/errata/eid8382
  */
-genPK : PrivateKey -> Address -> [n]Byte -> PublicKey
+genPK : PrivateKey -> [n]Byte -> Address -> PublicKey
 
 /**
  * Generate a signature from a private key and a message.
--- a/Primitive/Asymmetric/Signature/WOTS/Interface.cry
+++ b/Primitive/Asymmetric/Signature/WOTS/Interface.cry
@@ -72,6 +72,12 @@ type len_2 = lg2 (len_1 * (w - 1)) / lg2 w + 1
  * [RFC-8391] Section 3.1.1.
  */
 type len = len_1 + len_2
+/**
+ * This constraint comes from the fact that we set locations in `Address` to
+ * functions of `len` (in particular, `len / 2`). These locations are 32 bits
+ * in size, so we require that `width len` is no greater than 32 bits.
+ */
+type constraint (width len <= 32)
 
 /**
  * A byte is a sequence of 8 bits using big-endian representation.
--- a/Primitive/Asymmetric/Signature/WOTS/Interface.cry
+++ b/Primitive/Asymmetric/Signature/WOTS/Interface.cry
@@ -92,6 +92,24 @@ type Byte = [8]
 type Address = [8 * 4]Byte
 
 /**
+ * Address scheme for randomizing hash function calls in the OTS scheme.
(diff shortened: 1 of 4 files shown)


Primitive/Asymmetric/Signature/XMSS/Specification.cry
```diff
@@ -9,10 +9,450 @@
...
+     *
+     * The `treeHash` algorithm in [RFC-8391] does the computation in a
+     * depth-first manner, using a stack to store state. In this implementation,
+     * we use the breadth-first approach.
+     *
+     * [RFC-8391] Section 4.1.6, Algorithm 9.
+     */
+    treeHash : {s, t} (
+            // The start index.
+            fin s,
```
marsella, line 438: To convince myself of the correctness of this, I had to read a bit to confirm that the start index for a leaf is implicitly (since it's a leaf) at height 0. Consider adding a comment to that effect?

(I was trying to figure out why we'd never generate too many leaves).

Primitive/Asymmetric/Signature/XMSS/Specification.cry
```diff
@@ -9,10 +9,450 @@
...
+            s % (2^^t) == 0
+        ) => PrivateKey -> Address -> [2^^t]([n]Byte, Address)
+    treeHash_base SK ADRS = base where
+        base = [ (node, ADRS_3) where
+                 SEED = SK.seed
+                 ADRS_1 = setOTSAddress (setType ADRS 0) (`s + i)
+                 pk = WOTS::genPK (getWOTS_SK SK (`s + i)) SEED ADRS_1
+                 ADRS_2 = setLTreeAddress (setType ADRS_1 1) (`s + i)
+                 node = ltree pk SEED ADRS_2
+                 ADRS_3 = setTreeIndex (setTreeHeight (setType ADRS_2 2) 0) (i + `s)
```
marsella, line 385: suggestion: There are lots of different ways that `ADRS` is indicated in this file -- sometimes `ADRS''`, sometimes `ADRS_`, sometimes `ADRS_2`. It would be nice to try to consolidate the representations. Are you open to switching this to the tick notation? Or alternatively, switching all the instances of ticks to numbers?

Primitive/Asymmetric/Signature/WOTS/Interface.cry
```diff
@@ -72,19 +72,61 @@ type len_2 = lg2 (len_1 * (w - 1)) / lg2 w + 1
...
+ * reference implementation that accompanies the spec.
+ * @see https://github.com/XMSS/xmss-reference/blob/master/utils.c#L6
+ * [RFC-8391] Section 7.
+ *
+ * Essentially, we pad with zeros if needed (using `zext`), then `drop` any
+ * unnecessary padding (and potentially also drop some bits of the original
+ * integer).
+ */
+toByte : {y, m} (fin y, fin m) => [m] -> [y]Byte
+
```
marsella, line 105: I don't see where this is used, why is it in the public interface now?

marsella (commented):
Here are some comments and TODOs; I am not quite done reviewing but wanted to share some initial feedback.

Primitive/Asymmetric/Signature/WOTS/Interface.cry
amaloz: Ah, it's used in `sign`, which is not part of this PR. I can remove it and reintroduce it there.

Primitive/Asymmetric/Signature/XMSS/Specification.cry
amaloz: I prefer the `_N` notation, as, especially as `N` gets large, the ticks become harder to read in my opinion. So I'll switch to the `_N` notation if that's okay.

Primitive/Asymmetric/Signature/XMSS/Specification.cry
```diff
@@ -9,10 +9,275 @@
...
+}
+
+private
+    /**
+     * Take as input an XMSS private key `SK` and an integer `i` and output the
+     * `i`th WOTS+ private key.
+     *
+     * [RFC-8391] Section 4.1.3.
+     */
+    getWOTS_SK : PrivateKey -> Word -> WOTS::PrivateKey
```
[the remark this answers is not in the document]
amaloz: I believe the underlying XMSS implementation _will_ have an `i` fitting into a `Word`. See here:
> In addition to all XMSS parameters, an XMSS^MT system requires the
   number of tree layers d, specified as an integer value that divides h
   without remainder.  The same tree height h / d and the same
   Winternitz parameter w are used for all tree layers.

So XMSS will use a tree height of `h / d`, which in all parameter sets is less than 32.

Primitive/Asymmetric/Signature/XMSS/Specification.cry
marsella: Okay, here's the question I'm trying to get at: in your implementation, you always generate `2^^t` nodes. Imagine the 5-layer tree where you want the root node at `t=4` and there are 16 leaves at `t=0`. If I give you the start node that's the leaf node in the bottom left corner, you'll generate `2^t = 2^4 = 16` leaves and we're fine. If I give you the start node that's in a higher layer, you'll still generate 16 nodes, even though that layer is smaller than 16. In your implementation it looks like the `combine` function always combines all the nodes.

I think there's something wrong there.

amaloz: I believe `t` is the target node height _from the start index_. So in your second example `t = 4` really means the root node of the 5-layer tree (i.e., 4 from the level of `s`, which would be 5).

Take a look at the outer for loop in Algorithm 9: `for ( i = 0; i < 2^t; i++ )`. It always loops over all `2^t` elements and pushes something to the stack on each iteration.

marsella: Ah, okay. Yeah, the docs say "`s` is the leftmost leaf of a sub-tree of height `t`". I think that all tracks -- `t` is the height of the subtree, so from `s` to the "top"; `s` is a leaf in the subtree, so even if there are nodes beneath it in the "full" tree, we're not considering those when we talk about the height.

marsella (approved):
Looks good to me, thank you!

Issue #236, opened 2025-02-04
Review common code


marsella:
There are some files and functions in the `Common/` directory that may no longer be used after various outdated and duplicate specs have been removed from the repository.
- `mul_java` / `mod_arith` / `bv`
- `ByteArray` (can this be moved into `SHA.cry`?)
- `Field` / `Set` / `Morphism`
- `LUP`
- `pick`
- `utils`

Some of this is well-defined and documented but not used. Other stuff is not as well documented and the use cases have been lost over time.

- [ ] Define criteria for keeping things and document it in this issue
- [ ] Remove anything that should be removed
- [ ] Document any remaining functions and modules according to our standards, including docstrings on all properties

PR #292, opened 2025-04-23
Clean up ModDivZ

### LINKED ISSUE #291: Clean up PrimeField EC
I was exploring bringing RSA up to "gold standard", and came across some ~7 year old code to compute things like inverses in `Field`s and such. Presumably, things like the `prime` constraint and `recip` may not have existed at the time. Update the code so it is more idiomatic, and remove any files that have been obviated by improvements to the Cryptol language.

j-christensen1:
Closes #291


j-christensen1:
I have separated out commits in case we don't want all of them. The first just adds "modern" doc comments. The second replaces some functionality by Cryptol built-ins. The third removes a file and refactors it into `PFEC.cry`, the only place where it is imported.

marsella (approved):
Thanks, this is a good improvement! Related to #236 as well.

The only change I'd consider is moving this to `Common::utils` instead of `PFEC`, but I think fine to leave as-is for now. If we find ourselves writing other code that uses modular division we can extract this.

PR #294, opened 2025-04-25
Add verification scripts for aesIsCorrect

glguy:
These scripts are based off examples found in the SAW examples directory. Those examples had bitrotted because this specification had moved on. I think it would be useful for us to keep the script needed to verify the properties close to the spec so they can stay in sync now that it's updated.

### DIFF
--- a/Primitive/Symmetric/Cipher/Block/AES/Verifications/AES128.saw
+++ b/Primitive/Symmetric/Cipher/Block/AES/Verifications/AES128.saw
@@ -1,0 +1,35 @@
+/**
+ * Cryptol AES property verification.
+ *
+ * This module efficiently checks that decrypt is the inverse of encrypt.
+ *
+ * @copyright Galois Inc.
+ * @author Eric Mertens <emertens@galois.com>
+ */
+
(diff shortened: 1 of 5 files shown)


glguy:
With that last version each file takes about 2.5 seconds to run on my laptop

j-christensen1:
I think this looks good...the manual unfolding is perhaps the only "icky" part. I know there is an ongoing conversation about where SAW scripts should live, permanently, but I don't have a mature opinion on that.

marsella:
Since this is the first instance of SAW scripts being added into the repo as of recently, I would like to add a README to this directory to explain the role of this relative to the code fences. I'm happy to do it since I don't have direction beyond that yet. I will aim to add to his PR later this week.

glguy:
> I think this looks good...the manual unfolding is perhaps the only "icky" part.

I would love to see SAW evolve to not need that step!

marsella (approved):
Thanks for adding this! I added a brief README (please review to make sure it's accurate) and wrote up #297 to give us more automation of bit-rot-prevention.

PR #295, opened 2025-04-30
XTS mode implementation

glguy:
This implements the mode. There's still documentation and test vectors to do.


j-christensen1:
Going to add some test vectors and docs...

Primitive/Symmetric/Cipher/Block/Modes/XTS.cry
```diff
@@ -0,0 +1,254 @@
...
+      // The first `m` blocks are all 128 bits long.
+      // The final block, `Pm`, is between `0` and `127` bits long.
+      P_0_m2 # Pm1 # Pm = P
+
+      // Encrypt all but the final two blocks.
+      C_0_m2 = join [encryptBlock K1 T blk | T <- T_0_m1 | blk <- split`{M - 1} P_0_m2]
+
+      // Encrypt `P m - 1`, the final full block.
+      // Let `b` be the length of `Pm`.
+      // Steal the last `b` bits of the encryption
```
marsella, line 192: Should one of these be `128 - b`?

Primitive/Symmetric/Cipher/Block/Modes/XTS.cry
```diff
@@ -0,0 +1,254 @@
+/**
+ * Implementation of the XTS Tweakable Block Cipher mode.
+ *
```
marsella, line 3: What are your thoughts on adding an explanation up top re: which version of the standard we support? I'm inclined to do it so that users don't have to scroll through the spec to see that every function we implement is correct for both versions.

Primitive/Symmetric/Cipher/Block/Modes/XTS.cry
```diff
@@ -0,0 +1,254 @@
...
+ * the padding.
+ *
+ * Per section 5.3.1, the tweak is a 128 bit value.
+ *
+ * References:
+ *   [IEEE2007] Section 5.1, 5.3.1
+ *   [IEEE2018] Section 5.1, 5.3.1
+ */
+dataUnitSeqNoToIV : {n} (n <= 128) => [n] -> [128]
+dataUnitSeqNoToIV S = flipEndian (zext S)
```
marsella, line 92: This isn't used anywhere. Should it be? Or is that supposed to be handled by the calling application?

marsella (commented):
I can only leave a partial review because I don't have access to the spec at this time. Let me take this opportunity to voice my opposition to paywalled cryptography.

The explicit documentation throughout makes me relatively confident that this is correct, though. A few questions are inline.

Primitive/Symmetric/Cipher/Block/Modes/XTS.cry
glguy: I think the calling application chooses to use this if I want to use sequence numbers as the initialization vector.

I have it broken out because many test cases are given with a particular IV rather than a sequence number

Primitive/Symmetric/Cipher/Block/Modes/XTS.cry
glguy: I support adding a specific version here. I didn't do that because I didn't know which one we'd pick.

marsella, line 3: Second request up here: please note that this executable spec does not define the key scope, key export, or key archive described in Sections 6 and 7.

marsella (commented):
Here's a full review!

marsella (approved):
I'm satisfied with these changes, thanks Eric and John for all pitching in!

PR #305, opened 2025-06-03
SHA256 FFI implementation

amaloz:
This PR adds an FFI implementation of `SHA256::hashBytes`. It uses a public domain implementation of SHA256, and includes a test that validates that the FFI implementation outputs the same values as the Cryptol version.


marsella:
Thank you, those README updates look good to me!

PR #307, opened 2025-06-03
SHA3: Separate tests into particular instances.

j-christensen1:


Primitive/Keyless/Hash/SHA3/Tests/SHA3_224.cry
```diff
@@ -0,0 +1,76 @@
+/**
+ * Tests of the SHA3 hash functions.
```
marsella, line 2: suggestion (minor): update the headers for each file to specify which instantiation it's testing.

Issue #272, opened 2025-04-02
Add "Fallback Cache" on Default Branch


j-christensen1:
#268 adds the `check-docstrings` job to the CI. It is intended that the cache is scoped to each branch, with a 'fallback' cache pulled from the `master` branch if the cache does not exist (e.g., it is the first commit of a given PR). 

Unfortunately, the cache on `master` has not been built, so the first `check-docstrings` job of each PR is still expensive. The job has been configured to run on PRs, and so, hasn't been run on `master`.

- [x] Create a job that makes a fresh cache on `master` when a PR is merged.
- [x] Create a job that refreshes the cache on `master` so that it is not automatically evicted (the timeout is 7 days).
- [x] Consider using the `check-docstrings` job on `master` as a regression test (i.e., consider running the tests rather than just refreshing the cache).

j-christensen1:
Example: see [this workflow](https://github.com/GaloisInc/cryptol-specs/actions/runs/14229365257/job/39876416345) for an example of the behavior. The `master` key does not exist, so `check-docstrings` is run "fresh".

marsella:
Closed in #310.

PR #310, opened 2025-06-06
Try fixing cache behavior

marsella:
This makes two changes to the Cryptol projects cache, in pursuit of #272.
1. It runs the check-dockstrings job more frequently -- on a schedule and on push to main -- so that the master branch cache doesn't disappear or get out of date
2. It removes branch-specific caches. They were somehow not working well and made the config more complicated. I decided it's better to take the hit of rerunning changed tests multiple times per PR over not having a default cache and thus running all the tests on every PR.

This PR is brought to you by the 4-hour CI job for minor readme changes in #309.

I welcome suggestions on testing this other than just pushing to master and hoping for the best.

### DIFF
--- a/.github/workflows/check_docstrings.yml
+++ b/.github/workflows/check_docstrings.yml
@@ -5,13 +5,17 @@
 # @author John Christensen <jchristensen@galois.com>
 #
 
-# This job only runs on pull requests to the master branch.
-# We may expand this to other situations in the future.
 name: Check docstrings
+
 on:
+  # Run on a PR to master
   pull_request:
-    branches:
-      - master
+    branches: [master]
+  # Run on pushing to master
+  push:
+    branches: [master]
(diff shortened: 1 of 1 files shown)


j-christensen1 (commented):
Your explanation of the reasons for the changes seems plausible. 

I have not found a satisfactory testing method. I read a couple of different sources.

1. I was going to suggest a technique I was using on Gitlab of using a manually triggered scheduled job to get a cache made, but [this discussion seems to indicate this is perhaps not possible](https://github.com/orgs/community/discussions/16107).
2. I got excited about pushing a [cache from the command line](https://cli.github.com/manual/gh_cache), but it seems that this is not supported.
3. It seems that this can't be done through the [REST API either](https://docs.github.com/en/rest/actions/cache?apiVersion=2022-11-28).

PR #311, opened 2025-06-06
RSA Gold Standard

### LINKED ISSUE #293: Bring RSA primitives up to gold standard
RSA is defined in [RFC 8017](https://datatracker.ietf.org/doc/html/rfc8017). RSA is approved for key establishment (see e.g. [SP 800-175B rev1](https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-175Br1.pdf), Section 3.3.2.2 and [SP 800-56B rev2](https://csrc.nist.gov/pubs/sp/800/56/b/r2/final)) and for digital signatures (e.g. [SP 800-175B rev1](https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-175Br1.pdf), Section 3.3.1.4 and [FIPS 186-5](https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.186-5.pdf) Section 5).

We have many of the RSA primitives and instantiations defined, but they have not been brought up to our gold standard.

Other related reading:
- [RFC 4055](https://datatracker.ietf.org/doc/html/rfc4055#section-1.2) describes the usage of RSA encryption and signature schemes in X.509 PKI.
- [SP 800-78-5](https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-78-5.pdf) describes allowed use of RSA encryption and signature schemes for personal identity verification
- [IR 8547](https://nvlpubs.nist.gov/nistpubs/ir/2024/NIST.IR.8547.ipd.pdf) describes the eventual deprecation of RSA-based schemes
- [SP 800-57 Part 1 rev5](https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-57pt1r5.pdf) describes key management for many schemes including RSA
- [SP 800-89](https://nvlpubs.nist.gov/nistpubs/Legacy/SP/nistspecialpublication800-89.pdf) describes partial public key validation for RSA keys.

There is a lot of overlapping material. I think a good first step is to make sure the basic RSA primitives from Section 5 of RFC 8017 meet our gold standard, and also to decide whether the current organization (e.g. location of the primitives in the directory structure) makes sense when looking ahead to the schemes that will arise from those primitives. Right now we define 
- RSAEP and RSADP (Section 5.1) in [`Primitive/Asymmetric/Cipher/RSA.cry`](../tree/master/Primitive/Asymmetric/Cipher/RSA.cry)
- RSASP1 and RSAVP1 in  [`Primitive/Asymmetric/Signature/RSA.cry`](../tree/master/Primitive/Asymmetric/Signature/RSA.cry)

- [ ] Bring RSAEP/DP primitives up to gold standard
- [ ] Bring RSASP1/VP1 primitives up to gold standard
- [ ] Review directory structure and determine if it's appropriate
- [ ] Make sure there are citations to both the RSA definition in RFC 8017 and its approval for signatures in FIPS186-5 and for key establishment in SP 800-56Br1

j-christensen1:
Partially resolve #293

### DIFF
--- a/Primitive/Asymmetric/Cipher/RSA.cry
+++ b/Primitive/Asymmetric/Cipher/RSA.cry
@@ -51,6 +51,7 @@
  *
  *
  * @author John Christensen <jchristensen@galois.com>
+ * @author Ajay Kumar Eeralla
  * @copyright Galois, Inc.
  *
  */
--- a/Primitive/Asymmetric/Cipher/RSA.cry
+++ b/Primitive/Asymmetric/Cipher/RSA.cry
@@ -80,3 +80,32 @@ module Primitive::Asymmetric::Cipher::RSA where
  * 4. This specification does not address key generation. Keys must be generated
  *    in a particular way to conform to the standard.
  */
+
+/**
+ * Section 3.1 of [RFC 8017] specifies that the encryption exponent `e`
+ * is bounded by `3` and `n-1`.
+ */
+type constraint eValid n e = (e >= 3, e <= n - 1, fin n)
+
+/**
+ * Section 3.1 of [RFC 8017] specifies that `n` is the product of distinct
+ * odd primes.
+ *
+ * Deviation from RFC 8017:
+ * 1. This specification restricts the modulus to having two distinct factors.
+ * 2. `e` is required to satisfy a GCD condition that is tedious to express
+ *    in Cryptol's type system. Appeal to external validation will be required,
+ *    but is provided in this module.
+ *
+ * Non-conformance with NIST/FIPS:
+ * 1. FIPS 186-5, for instance, requires that the primes not only be distinct,
+ *    but also satisfy conditions on the absolute value of their difference.
+ *    These are propogated to the other standards, with some variations.
+ *    Consult each standard carefully for the particular requirements.
+ */
+type constraint nPartialValid p q n = (n == p * q, p != q, p >= 3, q >= 3, prime p, prime q)
+
+/**
+ * The conjunction of the previous constraints for syntactic convenience.
+ */
+type constraint RSAPubPartialValid p q n e = (nPartialValid p q n, eValid n e)
--- a/Primitive/Asymmetric/Cipher/RSA.cry
+++ b/Primitive/Asymmetric/Cipher/RSA.cry
@@ -1,100 +1,35 @@
-/* This module implements the data conversion and
-   cryptographic RSA primitives described in sections 4 and 5 of
-
-   PKCS #1: RSA Cryptography Specifications Version 2.2
-
-   See also:
-   https://tools.ietf.org/html/rfc8017
-
-   Copyright (c) 2018, Galois Inc.
-   www.cryptol.net
-*/
-
-
+/**
+ * This module implements a subset of the RSA encryption, decryption, signature,
+ * and signature verification primitives per their definitions in RFC 3447 (PKCS #1).
+ * These primitives are all modular exponentiations over a specially constructed modulus.
+ *
+ * The security of the RSA cryptographic primitives are subtle. A user of this module
+ * should note that conformance to RFC 3447 does not imply conformance to NIST or FIPS
+ * standards concerning integer factorization cryptography. We will reference these
+ * other standards when possible to illuminate the points of difference between them.
+ * We will also note future intent to implement the RSA primitives tailored to these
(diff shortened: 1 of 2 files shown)


j-christensen1:
@marsella @amaloz There is quite a lot of material about RSA, and a lot of potential approaches. For instance, one might start with key definitions from a particular standard, and define RSAEP as being only valid for keys conforming with the definition. On the other hand, the PKCS 1 / RFC 8017 pre-supposes the validity of the public/private key when it defines the primitives.

I think it is complicated enough that we should discuss the approach in a "code review" setting. Here, I have taken the following strategy:

1. Consider only bi-prime (as opposed to "multi-prime" RSA in the RFC) RSA, since this is required by all of the NIST standards. 
2. Use modular arithmetic types rather than integers, and document this deviation from the specification.
3. In the definitions of the primitives, do not assume the validity of the public/private key...assume that the key is validated elsewhere (as the RFC does).
4. Encode type constraints from the RFC in preference to NIST's constraints, with the idea that these modules can be augmented with NIST's definitions in further PRs. I feel like this is okay, because the RFC constraints, in my estimation, mostly encode functional correctness properties rather than security properties.

I will leave this marked as draft, but I nevertheless appreciate any feedback before I get too far down the rabbit hole. I've had multiple false starts with this PR want to keep the scope reasonable.

Primitive/Asymmetric/Cipher/RSA.cry
```diff
@@ -1,100 +1,293 @@
...
+ *
+ * `d` can be computed the extended Euclidean algorithm.
+ * We do specify the generation of `d` in this module.
+ *
+ * ```repl
+ * // 5 is its own inverse mod 12
+ * :prove dCorrect `{5, 7, 35, 5, 5}
+ * ```
+ */
+dCorrect :
```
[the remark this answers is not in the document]
marsella: We discussed this offline and decided that this implementation is operating under the assumption that keys are generated using the specified key generation procedure. Maybe add a comment about this failure mode.

Primitive/Asymmetric/Cipher/RSA.cry
```diff
@@ -1,100 +1,293 @@
...
+ * We will also note future intent to implement the RSA primitives tailored to these
+ * standards in other modules.
+ *
+ * This module is less general than RFC 8017. The author's have made the pragmatic
+ * decision to encode as many constraints specified by RFC 8017 in Cryptol's type system
+ * as possible. As a result, we allow RSA moduli with only two prime factors; the RFC
+ * allows any number of factors greater than or equal to two. The authors of this specification
+ * have made this decision based on the fact that NIST and FIPS standards allow only "bi-prime"
+ * RSA. Thus, the authors thought it a pragmatic middle ground between conformance to the RFC
+ * and ease of clear expression in Cryptol.
```
marsella, line 19: In other places I just say "we" when talking about decisions the authors of the cryptol specs have made.

Primitive/Asymmetric/Cipher/RSA.cry
```diff
@@ -1,100 +1,293 @@
...
+ *
```
marsella, line 20: todo: add an explicit note here about prime generation. Points to cover:
- Cryptol cannot express randomness generation so we can't easily describe correct key generation
- The specs require that primes are chosen uniformly at random e.g. with very little bias
- Many implementations try to optimize prime selection by using less-uniform methods. These can cause security vulnerabilities (you can link to some if you want) that are non-obvious
- This spec does not try to capture the range of failures that can happen with non-uniform prime generation.

You can use some warning emojis ⚠️ ⚠️ ⚠️ if you want.

marsella (changes requested):
We discussed this offline and I'm happy with the general approach here. The main changes I request are:
1. More explicit docs about how non-uniform prime generation can / will compromise the security of the whole system, and Cryptol cannot prevent it
2. More explicit constraints on the properties of the prime factors

For 2, in other places we have sort of separated the constraints from the general implementation -- see ECDSA's `Specification.cry` and `UnconstrainedSpecification.cry`. I don't mind if you want to do the same here, separating "pure" bi-prime RSA from "NIST-approved" RSA with various restrictions on the parameters.

marsella:
I'm closing this for now since it's not being actively worked on.

PR #312, opened 2025-06-11
TEST - DO NOT MERGE

marsella:
Testing to see whether the caching introduced in #310 worked. The behavior I expect to see is that the "Check docstrings" job runs quickly.


marsella:
This worked the way I expected, so I'm closing.

PR #315, opened 2025-06-23
A version of WOTS+ that uses the SHA256 FFI implementation.

amaloz:
The all-Cryptol version of WOTS+ is very slow, making testing it difficult. This version of WOTS+ uses the SHA256 FFI implementation to speed things up. Otherwise, it is exactly the same as the all-Cryptol version.

### DIFF
--- a/Primitive/Asymmetric/Signature/WOTS/Instantiations/WOTSP_SHA2_256_FFI.cry
+++ b/Primitive/Asymmetric/Signature/WOTS/Instantiations/WOTSP_SHA2_256_FFI.cry
@@ -1,0 +1,32 @@
+/**
+ * An instantiation of the WOTS+ one-time signature scheme using an FFI
+ * implementation of SHA256.
+ *
+ * These parameters and naming are drawn from [SP-800-208], Section 5 (Table 9
+ * and Section 5.1). The same parameters are also defined in [RFC-8391]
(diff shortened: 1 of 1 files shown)


Primitive/Asymmetric/Signature/WOTS/Instantiations/WOTSP_SHA2_256_FFI.cry
```diff
@@ -0,0 +1,32 @@
...
+ */
+module Primitive::Asymmetric::Signature::WOTS::Instantiations::WOTSP_SHA2_256_FFI =
+    Primitive::Asymmetric::Signature::WOTS::Specification where
+        import Primitive::Keyless::Hash::SHA2::Instantiations::SHA256_FFI as SHA256
+
+        type n' = 32
+        type w' = 16
+
+        F' KEY M = SHA256::hashBytes ((zero : [32][8]) # KEY # M)
+        PRF' KEY M = SHA256::hashBytes ((zero : [31][8]) # [(3 : [8])] # KEY # M)
```
marsella, line 32: I suppose we can't use the `toByte` function to instantiate the first parameters to these because it's not defined at this point. This is a little confusing but it's okay I think since they're constants.

marsella (approved):
Looks good, thank you!

PR #321, opened 2025-07-15
Fix names in Cryptol project config 

### LINKED ISSUE #320: Non-existent file `Primitive/Symmetric/Cipher/Block/AES/Tests/TestAES.cry` in `cryproject.toml`
The `cryproject.toml` lists this file:

https://github.com/GaloisInc/cryptol-specs/blob/bff67c85f27f2a11883b3bca49208a200394569d/cryproject.toml#L27

But this file does not actually exist. We should remove this part of the `cryproject.toml` file or replace it with something equivalent.

marsella:
These got dropped when addressing #238 and #277.

Closes #320

### DIFF
--- a/cryproject.toml
+++ b/cryproject.toml
@@ -24,11 +24,11 @@ modules = [
     "Primitive/Symmetric/Cipher/Block/AES/Instantiations/AES128.cry",
     "Primitive/Symmetric/Cipher/Block/AES/Instantiations/AES192.cry",
     "Primitive/Symmetric/Cipher/Block/AES/Instantiations/AES256.cry",
-    "Primitive/Symmetric/Cipher/Block/AES/Tests/TestAES.cry",
+    "Primitive/Symmetric/Cipher/Block/AES/Tests/*",
 
     # AES (counter mode)
     "Primitive/Symmetric/Cipher/Block/Modes/Instantiations/AES*_CTR.cry",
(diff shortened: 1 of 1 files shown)


cryproject.toml
```diff
@@ -24,11 +24,11 @@ modules = [
     "Primitive/Symmetric/Cipher/Block/AES/Instantiations/AES128.cry",
     "Primitive/Symmetric/Cipher/Block/AES/Instantiations/AES192.cry",
     "Primitive/Symmetric/Cipher/Block/AES/Instantiations/AES256.cry",
-    "Primitive/Symmetric/Cipher/Block/AES/Tests/TestAES.cry",
+    "Primitive/Symmetric/Cipher/Block/AES/Tests/*",
 
     # AES (counter mode)
     "Primitive/Symmetric/Cipher/Block/Modes/Instantiations/AES*_CTR.cry",
-    "Primitive/Symmetric/Cipher/Block/Modes/Tests/TestAES_CTR.cry",
+    "Primitive/Symmetric/Cipher/Block/Modes/Tests/TestAES*_CTR.cry",
```
amaloz, line 31: Looks like the names are of the form `AES{128|192|256}_CTR.cry`, so no `Test` prefix.

marsella: Ack. Thanks, fixed.

PR #324, opened 2025-08-05
Add whitespace check

marsella:
In pursuit of codifying some of the style guide elements in #5, I added a CI job to enforce invisible whitespace preferences. I also fixed all the aberrant files.


scripts/check_whitespace.sh
```diff
@@ -0,0 +1,59 @@
...
+function file_ends_with_newline() {
+    [[ $(tail -c1 "$1" | wc -l) -gt 0 ]]
+}
+
+function file_has_trailing_whitespace() {
+  lines=$(egrep -rnIH " +$" $1 | cut -f-2 -d ":")
+  if [ ! -z "$lines" ]; then
+    echo "$([[ -z "$tw_lines" ]] && echo "$lines" || echo $'\n'"$lines")"
+  fi
+}
```
sauclovian-g, line 17: This won't match hard tabs, which probably isn't what you want. To do that, you want `grep "[[:blank:]][[:blank:]]*$` or `egrep "[[:blank:]]+$"`.

Also, this seems unnecessarily complicated and likely slow compared to just feeding the whole list of filenames to `grep -l`...

marsella: Thank you! I read a lot of man pages and made the script better! This did catch one extra file with some tabs, and I think it's faster now for this step, at least (I didn't manage to avoid looping over files for newline-checking).

scripts/check_whitespace.sh
```diff
@@ -2,58 +2,58 @@
...
+  # `egrep` is used to ignore binary files (`-I`) and return the
+  # original file name (`-l`).
+  # The conditional:
+  # - checks the last byte of the file (`tail`),
+  # - counts the number of newlines (`wc`), and
+  # - fails if there are none (`gt`).
+  # If the conditional fails, we output the file name (`echo`).
+  for file in $(git ls-files | xargs egrep -Il ""); do
+    [[ $(tail -c1 "$file" | wc -l) -gt 0 ]] || printf '%s\n' "$file"
+  done
```
sauclovian-g, line 20: I have found the following slimy hack: `egrep -ILz "[[:space:]]$"` will print all files that don't end in a whitespace character, which includes newline because `-z` means to treat `\0` as the end-of-line character. This will not catch files whose last line ends in spaces and no newline, but the test for trailing whitespace will catch those.

I'm not sure how portable -z is though; I'll look into that...

scripts/check_whitespace.sh
sauclovian-g: It's depressing that stuff like this still has noticeable performance impact in 2025... but bash really is very slow :-(

kiniry: Rather than relying upon ad hoc scripts like this, perhaps we should consider moving to something like https://pre-commit.com/ in the medium term.  I am finding it very effective on another large project at Free & Fair.

scripts/check_whitespace.sh
marsella: Hm, that one's not working for me -- I'm getting the binary files back in because of `L`, and it's sometimes missing my manually-removed-newline in a non-binary file.

PR #328, opened 2025-08-20
Rewrite AES key wrap to match the NIST spec

### LINKED ISSUE #313: Add AES KeyWrap constraints from NIST spec
The AES KeyWrap implementation we have is based on [RFC 3394](https://datatracker.ietf.org/doc/rfc3394/), and similarly KeyWrapPadded is based on [RFC 5649](https://datatracker.ietf.org/doc/html/rfc5649). However, it doesn't specifically reference the NIST specification [SP 800-38F](https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-38F.pdf), which is "essentially equivalent" but has some differences (see Appendix B.1):
- It enforces an upper limit on plaintext length `n`
- It allows any approved 128-bit block cipher, not just AES (although there are no other ciphers approved for general use right now)

We should carefully review SP 800-38F and update our executable specification to match it more closely.

- [ ] Add plaintext length restriction
- [ ] Update documentation on the imported `CipherInterface`
- [ ] Updating references throughout to point to the appropriate sections in SP 800-38F
- [ ] Consider updating naming throughout to match the SP instead of the RFC
- [ ] Note and make any additional changes deemed appropriate

marsella:
Closes #313 

In pursuit of fixing some small bounds restrictions that we missed in the original gold-standard update of this spec, I ended up rewriting the whole thing. This now adheres to the names and the algorithm descriptions in [the NIST specification](https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-38F.pdf), which were largely equivalent to the RFC version ([RFC3394 for KeyWrap](https://datatracker.ietf.org/doc/rfc3394/) and [RFC5649 for key wrap with padding](https://datatracker.ietf.org/doc/rfc5649/)), but had relatively different presentations. Some high-level structural changes:
-  key wrap and key wrap with padding are now combined into one spec.
- the tests are drawn from [the ACVP server](https://github.com/usnistgov/ACVP-Server). I ended up separating them by top-level function (usually we do top-level algorithm) because they were so long.

Due to some renaming, the diff is not super helpful here -- the spec was written basically independently from the previous version (I pulled some details and individual lines) and later renamed it to use the original name, so the diff is basically incidental.

Review requests:
- Check for adherence to the spec -- naming / notation, ordering, code structure.
- Read the docs and confirm that they are coherent and adequately explain the differences from the spec
- I'm more interested in [presentation](https://github.com/GaloisInc/cryptol-specs/wiki/Reviewing-guidelines#questions-to-ask) than correctness. There are enough tests (400 😅) that I'm reasonably confident.


Primitive/Symmetric/Cipher/Block/Modes/AESKeyWrap.cry
```diff
@@ -18,241 +44,429 @@
...
+ * see also discussion in Appendix A.4.
  *
- *   Inputs:  Ciphertext, (n+1) 64-bit values {C0, C1, ..., Cn}, and
- *            Key, K (the KEK).
- *   Outputs: Plaintext, n 64-bit values {P0, P1, K, Pn}.
+ * This function is not approved for use independently of the KW_AE and KWP_AE
+ * functions. [SP-800-38F] Section 8.
+ */
+W : {n} (fin n, n >= 3, width (6 * (n - 1)) <= Semiblock)
+    => [C::KeySize] -> [n][Semiblock] -> [n * Semiblock]
```
marsella, line 244: Here and for `WInv`, I set it to actually take `[n][Semiblock]` instead of `[n * Semiblock]`, which means we call `split` at the call sites for each of those functions. Would it be preferable to `split` internally here?

Primitive/Symmetric/Cipher/Block/Modes/AESKeyWrap.cry
```diff
@@ -18,241 +44,429 @@
...
+ * 2. There are two additional failure checks, to make sure that `m` is correct
+ *    with respect to the values (`Plen`, `padlen`) derived from the ciphertext:
+ *    - `Plen` must be the same as `m`.
+ *    - `padlen == m %^ 8`. `padlen` is the number of octets of padding needed
+ *      to fill out the plaintext to the nearest semiblock (8 bytes) and `m` is
+ *      the number of octets in the plaintext. We use the ceiling modulus to
+ *      compute the expected amount of padding for `m`.
  *
- * [RFC-3394] Section 2.2.2.
+ * Thanks to Sean Weaver for assistance with formulating this definition.
```
weaversa, line 418: You're welcome. Thanks for your work creating a great specification.

Primitive/Symmetric/Cipher/Block/Modes/AESKeyWrap.cry
```diff
@@ -18,241 +44,429 @@
...
+        || (zero   != LSB`{8 * m %^ 64} S)  // Step 8.
+        || (Plen   != `m)                   // Cryptol-specific.
+        || (padlen != `(m %^ 8))            // Cryptol-specific.
+    then None
+    else Some P
+    where
+        S = KWP_AD_Step3 K C
+
+        // Step 5 - 6.
+        Plen = LSB`{32} (MSB`{64} S)
```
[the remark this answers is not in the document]
marsella: There is, thank you!

amaloz (approved):
Looks great!

Issue #331, opened 2025-09-11
Run all gold-standard docstrings in CI


marsella:
As evidenced by #329, it's a bit of a problem that we're only selectively running docstrings in CI -- in #268 we added the Suite B algorithms, but it would be better to run a larger set. Per our gold standard, every property should have at least one docstring repl command. Not all specs are gold standard, but the gold standard ones should all be checked in CI.

The current setup of CI is that each branch draws from the cache on `master`. When code is merged to master, that cache is updated. It may not be possible to add all the gold standard specs at once, if the additional docstrings take more than 4 hours to run. We might have to add them in several PRs, each of which increases the cache by a bit. Another option would be to have multiple CI jobs, each of which can run within the time limit, and multiple `cryproject.toml` files that divide up the algorithms in some (mostly arbitrary) way.

The things remaining to add that are gold standard include:
- [x] ML-KEM instantiations and tests
- [ ] We might add some more ECDSA instantiations beyond just the Suite B approved ones
- [x] ML-DSA instantiations and tests
- [x] SHA1 spec and tests
- [x] SHA3 instantiations and tests
- [x] Modes of operations, instantiated for AES: CBC, CFB, XTS, and key wrap. And their tests
- [x] Simon and Speck ciphers
- [x] HMAC instantiations and tests

I need to double-check XMSS, which I think is gold standard, although not entirely implemented (the XMSS-MT algorithm isn't written yet). I also need to confirm that docstrings can handle FFI instantiations of things, in case we want to also test the FFI version of SHA2.

PR #332, opened 2025-09-11
Add gold-standard specs to CI

marsella:
This updates the `cryproject.toml` file with additional hash-related specs: SHA1, SHA3, and HMAC. **Update:** I also added AES modes of operation (excluding XTS and key wrap) and Simon and Speck. 

Partially addresses, but does not complete, #331.

### DIFF
--- a/Primitive/Asymmetric/Signature/SphincsPlus/3.1/specification.tex
+++ b/Primitive/Asymmetric/Signature/SphincsPlus/3.1/specification.tex
@@ -2352,7 +2352,7 @@ generation.
     type h : #
     type constraint (fin h)
     type d : #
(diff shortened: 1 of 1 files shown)


marsella:
An earlier CI run failed because of [cryptol#1903](https://github.com/GaloisInc/cryptol/issues/1903). Rather than modify the files here, I removed them from the `cryproject.toml` file and will keep #331 open until that issue is fixed.

I rebased to pull in the changes from #330, which also contributed to the failure of an earlier pipeline run.

Finally, since the run only takes 1 hour now and we have 4 to play with, I'm continuing to add additional specs, including the modes of operation for AES and the Simon and Speck ciphers.

Ra1issa:
lgmt

Ra1issa (approved):

Issue #334, opened 2025-09-30
`Primitive::Asymmetric::Signature::ECDSA::Tests::ECDSA_P{224,256,384}_SHA{224,256,384}` fail `:check`


RyanGlScott:
Each of `Primitive::Asymmetric::Signature::ECDSA::Tests::ECDSA_P224_SHA224`, `Primitive::Asymmetric::Signature::ECDSA::Tests::ECDSA_P256_SHA256`, and `Primitive::Asymmetric::Signature::ECDSA::Tests::ECDSA_P384_SHA384` fail if you load them into the Cryptol REPL and run `:check`. For instance:

```
$ ~/Software/cryptol-3.3.0/bin/cryptol Primitive/Asymmetric/Signature/ECDSA/Tests/ECDSA_P224_SHA224.cry -c ":check"
Loading module Cryptol
Loading interface module Common::EC::ECInterface
Loading interface module Primitive::Keyless::Hash::HashInterface
Loading module Primitive::Asymmetric::Signature::ECDSA::UnconstrainedSpec
Loading module Primitive::Asymmetric::Signature::ECDSA::Specification
Loading module Primitive::Keyless::Hash::SHA2::Instantiations::SHA224__where
Loading interface module Primitive::Keyless::Hash::SHA2::Specification__parameter
Loading module Primitive::Keyless::Hash::SHA2::Specification
Loading module Primitive::Keyless::Hash::SHA2::Instantiations::SHA224
Loading module Common::EC::PrimeField::Instantiations::P224__where
Loading module Common::utils
Loading interface module Common::EC::PrimeField::PFEC__parameter
Loading module Common::EC::PrimeField::PFEC
Loading module Common::EC::PrimeField::Instantiations::P224
Loading module Primitive::Asymmetric::Signature::ECDSA::Instantiations::ECDSA_P224_SHA224
Loading module Primitive::Asymmetric::Signature::ECDSA::Tests::ECDSA_P224_SHA224
property checkSignature Showing a specific instance of polymorphic result:
  * Using '0' for 1st type argument of 'Primitive::Asymmetric::Signature::ECDSA::Instantiations::ECDSA_P224_SHA224::sign'
Using random testing.
Testing... Counterexample
checkSignature
  19878665429645173130939544429917673631368586938530619164285815298784
  0x0
  17864806397770356865377037310229800501440565370158022435387214530648
  20547304272684529263248073456472189910526508729095380119333625863517
[...]

PR #335, opened 2025-09-30
`ECDSA`: don't mark `checkSignature` as a property

RyanGlScott:
`checkSignature` would be run with `:check` (since it was marked as a `property`) but then fail. `checkSignature` isn't meant to be used as a standalone property anyway, but rather as a building block for constructing other properties (e.g., `sampleVector`). Let's remove the `property` keyword to reflect this and fix `:check`.

Fixes https://github.com/GaloisInc/cryptol-specs/issues/334.


marsella (approved):
Thanks for catching this!

PR #336, opened 2025-10-01
Add more specs to CI

marsella:
This adds several more existing gold-standard specs to the `cryproject.toml` file that is run by CI. In particular:
- XTS and AES key wrap modes have been unblocked with GaloisInc/cryptol#1903
- ML-KEM is fully added
- Most of the instantiations for ML-DSA are added. [A previous CI run timed out](https://github.com/GaloisInc/cryptol-specs/actions/runs/17743986804/job/50424627399) while running the most secure of these, ML-DSA-87, so I skipped it here to ensure we can finish adding the rest of the specs.

Reviewer requests: Make sure I typed all the file names correctly, and that CI runs successfully.

Mostly addresses, but does not complete #331


Ra1issa (approved):
Reviewed this by checking that all these files exist. And they do! lgtm

Issue #136, opened 2024-09-05
Expand ECDH test vectors


marsella:
Right now we have the CAVP test vectors to demonstrate the correct behavior for ECDH with P-256. We can expand this in two ways. First, we don't have any negative test vectors (e.g. to show that things aren't trivially correct / even invalid keys are approved). It might be good to think about whether there's a principled way to do this or if we should just mash some of the valid vectors to get invalid ones. Second, we could instantiate ECDH for more curves. [NIST SP 800-186](https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-186.pdf) allows all the prime-field curves to be used for EC key agreement (section 3.1.2, Table 2), so we might instantiate all of those and pull in all the corresponding CAVP test vectors.

- [ ] Add some negative test cases for each method in ecdh
- [ ] Add additional P-curves (P-224, P-384, P521) and CAVP test vectors

PR #340, opened 2025-11-06
Add ECDH instantiations for P-224 and P-521

RyanGlScott:
This checks off two boxes in #136.


RyanGlScott:
Question: should I update [this row](https://github.com/GaloisInc/cryptol-specs/blob/e2412b4a69f00ef76926278fb259bd16d0229e21/README.md#L48) in the table in the `README` to mention these new instantiations?

marsella (approved):
Looks good, thanks for adding these!

I spot-checked the test vectors and they look good. Conveniently, they were automatically pulled into the CI due to some convenient wildcard action.

PR #341, opened 2025-12-03
ci: update CI to handle partially-cached docstrings

marsella:
This removes ML-DSA-87 from CI in an attempt to rebuild the cache, which was invalidated due to a change in the Cryptol prelude, and which takes longer than the maximum CI runtime of 4 hours to rebuild in its entirety.

### DIFF
--- a/.github/workflows/check_docstrings.yml
+++ b/.github/workflows/check_docstrings.yml
@@ -65,22 +65,28 @@ jobs:
 
       # Run the check-docstrings script, regardless of cache load.
       - name: "Run the check-docstrings script."
+        id: checkscript
         run: |
           git config --global --add safe.directory '*'
           echo "Check which caches exist, for sanity."
           gh cache list
 
           bash scripts/check_docstrings.sh .
+        # Timeout is slightly less than 6 hours to allow saving a cache in the case of timeout.
+        # The following steps don't run if this step failed, but do run if it was cancelled
+        # (e.g. due to hitting this timeout).
+        timeout-minutes: 355
+        continue-on-error: true
 
       # Caches can only be uploaded once. If we're on master and the cache already exists,
       # we need to delete it...
       - name: "If we're on master, delete the cache"
-        if: github.ref == 'refs/heads/master' && steps.load-cache.outputs.cache-hit == 'true'
+        if: github.ref == 'refs/heads/master' && steps.load-cache.outputs.cache-hit == 'true' && steps.checkscript.outcome != 'failure'
         run: gh cache delete ${{ env.CRYPTOL_CACHE_KEY }}
 
       # ...before we upload it. Do this regardless of whether it previously existed.
       - id: save-cache
-        if: github.ref == 'refs/heads/master'
+        if: github.ref == 'refs/heads/master' && steps.checkscript.outcome != 'failure'
         uses: actions/cache/save@v4
         with:
           path: ${{ env.CRYPTOL_CACHE }}
--- a/.github/workflows/check_docstrings.yml
+++ b/.github/workflows/check_docstrings.yml
@@ -91,3 +91,10 @@ jobs:
         with:
           path: ${{ env.CRYPTOL_CACHE }}
           key: ${{ env.CRYPTOL_CACHE_KEY }}
+
+      # After potentially uploading the cache, we fail according to the checkscript step.
+      - name: "Check whether the script failed"
+        if: steps.checkscript.outcome != 'success'
+        run: |
+          echo "Docstrings script failed with ${{ steps.checkscript.outcome }}"
+          exit 1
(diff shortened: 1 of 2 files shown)


marsella:
A few progress notes:
- We identified that the cache break was due to a change in the Cryptol prelude, which invalidates every other cache entry.
- Removing ML-DSA-87 was not enough to reduce the docstrings runtime to less than 6 hours

I added a commit that aims to save the cache in cases where it was cancelled (e.g. due to timeout). Thanks to GaloisInc/cryptol#1912, the cache should now be saved incrementally, even if it doesn't finish entirely. I might have to modify the way we call the check-docstrings command to use the correct command-line arguments, which changed in that fix.

marsella:
I'm not going to change the command that checks the docstrings; it's correct as-is.

I'm not sure how to test this outside of pushing it to main and seeing if it works. 

Is it worth printing some debug information (like, printing the cache file) to more easily track if it's working as intended?

.github/workflows/check_docstrings.yml
```diff
@@ -65,23 +65,36 @@ jobs:
...
       # we need to delete it...
       - name: "If we're on master, delete the cache"
-        if: github.ref == 'refs/heads/master' && steps.load-cache.outputs.cache-hit == 'true'
+        if: github.ref == 'refs/heads/master' && steps.load-cache.outputs.cache-hit == 'true' && steps.checkscript.outcome != 'failure'
         run: gh cache delete ${{ env.CRYPTOL_CACHE_KEY }}
 
       # ...before we upload it. Do this regardless of whether it previously existed.
       - id: save-cache
-        if: github.ref == 'refs/heads/master'
+        if: github.ref == 'refs/heads/master' && steps.checkscript.outcome != 'failure'
```
aschwerdfeger-galois, line 89: **Question:** Does GitHub have a separate outcome for timed-out steps to distinguish timeouts from non-timeout failures? If not, it looks like the cache would not be updated here in case of a timeout.

.github/workflows/check_docstrings.yml
```diff
@@ -65,23 +65,36 @@ jobs:
...
       # Run the check-docstrings script, regardless of cache load.
       - name: "Run the check-docstrings script."
+        id: checkscript
         run: |
           git config --global --add safe.directory '*'
           echo "Check which caches exist, for sanity."
           gh cache list
 
           bash scripts/check_docstrings.sh .
+        # Timeout is slightly less than 6 hours to allow saving a cache in the case of timeout.
```
aschwerdfeger-galois, line 75: **Question:** Is it possible that the abrupt termination of this step, in case of a timeout, would leave the cache directory in any kind of inconsistent state that would cause problems if reused later?

aschwerdfeger-galois (commented):
Left a couple of questions in the code.

.github/workflows/check_docstrings.yml
marsella: [The options](https://docs.github.com/en/actions/reference/workflows-and-actions/contexts#steps-context) are success, failure, cancelled, or skipped. Timeouts lead to cancellation [as far as I can tell](https://github.com/GaloisInc/cryptol-specs/actions/runs/20211255150). This isn't perfect because you can also manually cancel a job, but I couldn't find a way to check for time-outs specifically.

.github/workflows/check_docstrings.yml
marsella: @yav updated the writing step of the cache update to be more atomic, so instead of writing live to the cache file, it writes to a temporary file, then `mv`s it into the cache one it completed successfully. I'm hopeful this will be enough to avoid a half-written or otherwise invalid cache file. I think the project script will ignore the temporary files if they exist, so it's fine if they end up half-formed.

Those were the only failure cases I could think of, since the cache is just a single file. Is there anything else that might be a concern?

aschwerdfeger-galois: That sounds like adequate protection against fugue states.

.github/workflows/check_docstrings.yml
aschwerdfeger-galois: In that case, the only problem scenario I can think of is if a job is canceled or times out on its way to failure and something unwanted gets in the cache from that, but it looks like this is also taken care of by the cache being populated atomically.

aschwerdfeger-galois (approved):

kiniry:
FWIW, I have been using the [act](https://github.com/nektos/act) tool in order to run CI actions locally, including those that are representative of "push to main and see if it works".  We might consider adding it to the build system.

.github/workflows/check_docstrings.yml
marsella: Unfortunately, timeout results in a `failure` outcome, not a `cancelled` one. I'm fixing this in a separate branch -- should've tested it before merging 😬

marsella: Fixed in #343

PR #345, opened 2026-03-16
Rearrange imports of parameter blocks and submodules

### LINKED ISSUE #344: Reorder submodule imports and parameter blocks
Now that https://github.com/GaloisInc/cryptol/pull/2002 is merged, Cryptol has changed the scope in which some parameters are defined. Submodules, imports of submodules, and fields in `parameter` blocks are now only in scope after they're defined.

- [ ] Rearrange all the modules that no longer build

marsella:
Closes #344.

I think this might be a little tricky CI-wise: the Cryptol update made some of these files invalid / unable to load. I think it also updated the Cryptol prelude, which invalidates the cache. Recently, we fixed CI to be able to handle a Cryptol prelude update on main by rebuilding the cache over a couple of CI runs. However, this assumed that we could actually run on main. Briefly:
- main's cache is broken
- This branch needs main's cache to run the docstrings within the time limit
- main's cache won't get fixed until this branch is merged

I checked the docstrings on some of the affected files locally, using Cryptol nightly (version 3.5.0.99 ([d3027bc](https://github.com/GaloisInc/cryptol/commit/d3027bc5230c531046756fff78760b8b8147de95))) on the instantiation and tests for: ML-KEM 512, SHA3-256, SHAKE128, and ML-DSA44, so I think this will be ultimately pass.

### DIFF
--- a/.github/workflows/check_docstrings.yml
+++ b/.github/workflows/check_docstrings.yml
@@ -57,13 +57,13 @@ jobs:
 
       # Check out the code.
       - id: checkout
-        uses: actions/checkout@v4
+        uses: actions/checkout@v5
         with:
           fetch-depth: 0
 
       # Pull the default cache
       - id: load-cache
-        uses: actions/cache/restore@v4
+        uses: actions/cache/restore@v5
         with:
           path: ${{ env.CRYPTOL_CACHE }}
           key: ${{ env.CRYPTOL_CACHE_KEY }}
@@ -100,7 +100,7 @@ jobs:
       # ...before we upload it. Do this regardless of whether it previously existed.
       - id: save-cache
         if: github.ref == 'refs/heads/master'
-        uses: actions/cache/save@v4
+        uses: actions/cache/save@v5
         with:
           path: ${{ env.CRYPTOL_CACHE }}
           key: ${{ env.CRYPTOL_CACHE_KEY }}
--- a/.github/workflows/check_docstrings.yml
+++ b/.github/workflows/check_docstrings.yml
@@ -33,6 +33,11 @@ jobs:
       # There's one cache for the whole repo. It's associated with the `master` branch.
       CRYPTOL_CACHE_KEY: cryptol-specs-cache
 
+      # Backup key, if we need to rebuild the cache from scratch on a branch.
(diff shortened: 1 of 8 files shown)


marsella:
I ended up adding a branch cache feature that runs on PRs with the `use-branch-cache` label. This seems to have worked; that cache is getting updated.

The branch cache is shared amongst all PRs but we've never had a situation where that would matter, so I'm not going to specialize it for now. It also doesn't get deleted after the PR is closed / merged, so it might be out of date on subsequent uses. I think this is fine because it's mainly used for scenarios like this, where the entire cache is invalid anyways.

.github/workflows/check_docstrings.yml
```diff
@@ -63,6 +68,14 @@ jobs:
           path: ${{ env.CRYPTOL_CACHE }}
           key: ${{ env.CRYPTOL_CACHE_KEY }}
 
+      # Use branch cache
+      - id: load-branch-cache
+        uses: actions/cache/restore@v5
```
aschwerdfeger-galois, line 73: **Suggestion:** Handle the master and branch caches with the same version of the cache actions.

.github/workflows/check_docstrings.yml
```diff
@@ -92,6 +105,20 @@ jobs:
           path: ${{ env.CRYPTOL_CACHE }}
           key: ${{ env.CRYPTOL_CACHE_KEY }}
 
+      # If we used a branch cache, delete and re-upload it.
+      - name: "If we used a branch cache, delete it"
+        if: |
+          github.ref != 'refs/heads/master' &&
```
aschwerdfeger-galois: **Question:** Why is this condition only present on this deletion step, and not on the save / restore steps?

marsella: Ah, that's an artifact of my thought process. At first I was going to run on all PRs, but then realized I could save some potential conflict by only using the cache when we've specifically labeled the PR to do so. 

There's probably a weird scenario where we're making a PR to merge main into another branch and also label that PR where we might end up saving both caches. But I don't even think that's a problem, necessarily.

I'll remove it.

aschwerdfeger-galois (approved):

PR #346, opened 2026-03-20
Update and simplify CI

marsella:
This addresses two maintenance issues for the CI:
1. Replaces superlinter, and its use of a highly-permissioned `GITHUB_TOKEN`, with direct use of the editorconfig tool, which is the only one we're using out of the many provided by superlinter. This removes the pretty-printed checkbox from the action, which seems fine to me.
2. Updates all the checkout actions to the latest version.

[Here's a sample run](https://github.com/GaloisInc/cryptol-specs/actions/runs/23349380464/job/67923511625) where I introduced poorly-linted code.

Closes https://github.com/GaloisInc/cryptol-specs/security/dependabot/1.

### DIFF
--- a/.github/workflows/check_docstrings.yml
+++ b/.github/workflows/check_docstrings.yml
@@ -57,7 +57,7 @@ jobs:
 
       # Check out the code.
       - id: checkout
-        uses: actions/checkout@v5
+        uses: actions/checkout@v6
         with:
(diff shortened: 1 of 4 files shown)


aschwerdfeger-galois (approved):
If the CI says this is good, so do I.

PR #350, opened 2026-06-11
Add AES Key Wrap instantiations for 128 and 192

RyanGlScott:
These complement the existing `AES256_KeyWrap` instantiation and tests.


Primitive/Symmetric/Cipher/Block/Modes/Instantiations/AES128_KeyWrap.cry
```diff
@@ -0,0 +1,11 @@
+/*
+ * Instantiate AES key wrap and AES key wrap with padding for AES 192.
+ *
+ * @copyright Galois, Inc.
+ * @author Marcella Hastings <marcella@galois.com>
```
marsella: suggestion: You might change the author name in the instantiations. The git history will out me as a plagiarist.

Primitive/Symmetric/Cipher/Block/Modes/Instantiations/AES128_KeyWrap.cry
```diff
@@ -0,0 +1,11 @@
+/*
+ * Instantiate AES key wrap and AES key wrap with padding for AES 192.
```
marsella: todo: fix docs.
```suggestion
 * Instantiate AES key wrap and AES key wrap with padding for AES 128.
```

marsella (approved):
Looks good to me, modulo one tiny doc change. I ran all the docstrings manually and they worked.

Primitive/Symmetric/Cipher/Block/Modes/Instantiations/AES128_KeyWrap.cry
RyanGlScott: Fair enough. I originally attributed you because I only contributed derivative work, but I'm happy to take all the credit :)

PR #354, opened 2026-08-06
Fix overlapping numeric constraint guards in Speck.

### LINKED ISSUE #353: Overlapping Numeric Constraint Guards In Speck

j-christensen1:
Closes #353 

The overlapping numeric constraint guards are evaluated to the correct values, so the Cryptol spec is still correct for any valid instantiation. That being said, it reads incorrectly.

Thanks @sauclovian-g


marsella (approved):
Looks fine to me, thanks!

PR #356, opened 2026-08-25
FALCON: Modify property to ignore NaNs 

### LINKED ISSUE #352: FALCON: `FFTInvFFT` property does not hold
FALCON's [`FFTInvFFT` property](https://github.com/GaloisInc/cryptol-specs/blob/82023f4bee71e8d6f00e24e926b04e33cf2aacd6/Primitive/Asymmetric/Signature/FALCON/1.2/spec2.tex#L818-L819) does not hold in general, as can be seen if you try to `:check` it:

```
$ cryptol Primitive/Asymmetric/Signature/FALCON/1.2/falcon_512.cry
<snip>
falcon_512> :check FFTInvFFT
Using random testing.
Counterexample
FFTInvFFT
  [(0x6.1f1968a8ca57cp444, fpNaN), (0.0, fpNegInf),
   (-0x5.034e10cc83e24p-276, 0xd.f4abfd174a6e8p-348),
   (0.0, fpPosInf)]
  = False
```

The problem `fpNaN`, which is not equal to itself according to `(==)`. The [Falcon spec](https://falcon-sign.info/falcon.pdf) has this to say about NaN:

> The exponent _e_ uses 11 bits, but its range covers only 2046 values, not 2048. The two extra possible values for that field encode special cases:
>
> * The value zero. IEEE 754 has two zeros, that differ by the sign bit.
> * Subnormals: they use the minimum value for the exponent (−1022) but the implicit top bit of the mantissa is 0 instead of 1.
> * Infinites (positive and negative).
> * Erroneous values, known as NaN (Not a Number).
>
> Apart from zero, Falcon does not exercise these special cases; exponents remain relatively close to zero; no infinite or NaN is obtained.

As such, I think we can safely ignore NaN. One possible way to repair this property would be to filter out `FFT` values which contain NaN, e.g.,

```
  CHasNaN : C -> Bit
  CHasNaN x = fpIsNaN x.0 \/ fpIsNaN x.1

  FFTHasNaN : {n} (fin n) => FFT n -> Bit
  FFTHasNaN = any CHasNaN

  FFTInvFFT : (FFT 4) -> Bit
  property FFTInvFFT f = ~ (FFTHasNaN f) ==> mergefft`{2,4} (splitfft`{2,4} f) == f
```

Another option would be to use Cryptol's `(=.=)` function (which implements a "logical" equality where `fpNaN =.= fpNaN` is true) instead of `(==)`:

```
  CLogicalEq : C -> C -> Bit
  CLogicalEq (x0, x1) (y0, y1) = (x0 =.= y0) /\ (x1 =.= y1)

  FFTLogicalEq : {n} (fin n) => FFT n -> FFT n -> Bit
  FFTLogicalEq f1 f2 = and (zipWith CLogicalEq f1 f2)

  FFTInvFFT : (FFT 4) -> Bit
  property FFTInvFFT f = FFTLogicalEq (mergefft`{2,4} (splitfft`{2,4} f)) f
```

marsella:
Closes #352 

Thanks for flagging this, Ryan.

### DIFF
--- a/Primitive/Asymmetric/Signature/FALCON/1.2/spec2.tex
+++ b/Primitive/Asymmetric/Signature/FALCON/1.2/spec2.tex
@@ -815,8 +815,26 @@ The \falcon public key \pk corresponding to the private key $\sk =
  It is immediate that \polsplit and \polmerge are inverses of each other, and equivalently \splitfft and \mergefft are inverses of each other. Just as for \splitfft, readers familiar with the Fourier transform can observe that \mergefft is a step of the fast Fourier transform: it is the reconstruction step which from two small \fft's computes a larger \fft.
 
 \begin{code}
+  // Check if a complex number is erroneous (NaN -- Not a Number).
+  CHasNaN : C -> Bit
+  CHasNaN x = fpIsNaN x.0 \/ fpIsNaN x.1
+
+  // Check if an FFT value has any invalid entries.
+  FFTHasNaN : {n} (fin n) => FFT n -> Bit
+  FFTHasNaN = any CHasNaN
+
+  /**
+   * The split and merge functions are inverses of each other.
+   * Section 4.1 claims that FALCON does not exercise "special case" FFT values,
+   * including infinites or NaNs; NaN is explicitly excluded here to achieve
+   * `:check`ability.
+   *
+   * ```repl
+   * :check FFTInvFFT
+   * ```
+   */
   FFTInvFFT : (FFT 4) -> Bit
-  property FFTInvFFT f = mergefft`{2,4} (splitfft`{2,4} f) == f
+  property FFTInvFFT f = ~ (FFTHasNaN f) ==> mergefft`{2,4} (splitfft`{2,4} f) == f
 \end{code}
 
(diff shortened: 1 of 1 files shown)


Primitive/Asymmetric/Signature/FALCON/1.2/spec2.tex
```diff
@@ -815,8 +815,26 @@ \subsection{Public key}
...
+  CHasNaN x = fpIsNaN x.0 \/ fpIsNaN x.1
+
+  // Check if an FFT value has any invalid entries.
+  FFTHasNaN : {n} (fin n) => FFT n -> Bit
+  FFTHasNaN = any CHasNaN
+
+  /**
+   * The split and merge functions are inverses of each other.
+   * Section 4.1 claims that FALCON does not exercise "special case" FFT values,
+   * Including infinites or NaNs; NaN is explicitly excluded here to achieve
```
RyanGlScott:
```suggestion
   * Section 4.1 claims that FALCON does not exercise "special case" FFT values,
   * including infinites or NaNs; NaN is explicitly excluded here to achieve
```

RyanGlScott (approved):

PR #358, opened 2026-09-02
`falcon`: Directly instantiate `spec2` #357

### LINKED ISSUE #357: FALCON: Make `falcon_{512,1024}` instantiate `spec2` directly rather than defining instantiations as submodules
`cryptol-specs`' FALCON implementation provides "instantiations" of the spec ([`spec2`](https://github.com/GaloisInc/cryptol-specs/blob/9bb642049b3cb9175a56e91c839138b963d2c9b9/Primitive/Asymmetric/Signature/FALCON/1.2/spec2.tex)) by way of the [`falcon_512`](https://github.com/GaloisInc/cryptol-specs/blob/9bb642049b3cb9175a56e91c839138b963d2c9b9/Primitive/Asymmetric/Signature/FALCON/1.2/falcon_512.cry) and [`falcon_1024`](https://github.com/GaloisInc/cryptol-specs/blob/9bb642049b3cb9175a56e91c839138b963d2c9b9/Primitive/Asymmetric/Signature/FALCON/1.2/falcon_1024.cry) modules. I use "instantiations" in scare quotes, however, because `falcon_{512,1024}` are defined in a somewhat atypical way. Rather than instantiating `spec2` directly, they define submodules that perform the instantiations. For example, in `falcon_512` we have:

https://github.com/GaloisInc/cryptol-specs/blob/9bb642049b3cb9175a56e91c839138b963d2c9b9/Primitive/Asymmetric/Signature/FALCON/1.2/falcon_512.cry#L5-L8

The `falcon_512` module doesn't define anything besides the `falcon_inst` submodule, so anyone who imports `falcon_512` would then need to import the `falcon_inst` submodule in order to do anything with it. This is possible, but rather indirect.

I propose that we cut out this extra step and instead make `falcon_{512,1024}` instantiate `spec2` directly. That is, do something like:

```diff
diff --git a/Primitive/Asymmetric/Signature/FALCON/1.2/falcon_1024.cry b/Primitive/Asymmetric/Signature/FALCON/1.2/falcon_1024.cry
index 40dff6e..d9b80d6 100644
--- a/Primitive/Asymmetric/Signature/FALCON/1.2/falcon_1024.cry
+++ b/Primitive/Asymmetric/Signature/FALCON/1.2/falcon_1024.cry
@@ -2,9 +2,7 @@
  * @copyright Galois, Inc
  * @author Marios Georgiou <marios@galois.com>
  */
-module falcon_1024 where
-
-submodule falcon_inst = spec2
+module falcon_1024 = spec2
     where
     type _k = 10
     type _n = 1024
@@ -1044,5 +1042,3 @@ submodule falcon_inst = spec2
         (0.9999247018391445, -0.012271538285720572),
         (0.9999811752826011, -0.006135884649154477)
     ]
-
-import submodule falcon_inst
diff --git a/Primitive/Asymmetric/Signature/FALCON/1.2/falcon_512.cry b/Primitive/Asymmetric/Signature/FALCON/1.2/falcon_512.cry
index 9682024..9b8881e 100644
--- a/Primitive/Asymmetric/Signature/FALCON/1.2/falcon_512.cry
+++ b/Primitive/Asymmetric/Signature/FALCON/1.2/falcon_512.cry
@@ -2,9 +2,7 @@
  * @copyright Galois, Inc
  * @author Marios Georgiou <marios@galois.com>
  */
-module falcon_512 where
-
-submodule falcon_inst = spec2
+module falcon_512 = spec2
     where
     type _k = 9
     type _n = 512
@@ -532,5 +530,3 @@ submodule falcon_inst = spec2
         (0.9996988186962042, -0.02454122852291245),
         (0.9999247018391445, -0.012271538285720572)
     ]
-
-import submodule falcon_inst
```

RyanGlScott:
This makes `falcon_{512,1024}` directly instantiate `spec2` rather than instantiating `spec2` indirectly via a submodule.

Fixes #357.

### DIFF
--- a/Primitive/Asymmetric/Signature/FALCON/1.2/falcon_1024.cry
+++ b/Primitive/Asymmetric/Signature/FALCON/1.2/falcon_1024.cry
@@ -2,10 +2,7 @@
  * @copyright Galois, Inc
  * @author Marios Georgiou <marios@galois.com>
  */
-module falcon_1024 where
-
-submodule falcon_inst = spec2
-    where
+module falcon_1024 = spec2 where
     type _k = 10
     type _n = 1024
     type pkbytelen = 1793
@@ -1044,5 +1041,3 @@ submodule falcon_inst = spec2
         (0.9999247018391445, -0.012271538285720572),
         (0.9999811752826011, -0.006135884649154477)
     ]
-
-import submodule falcon_inst
--- a/Primitive/Asymmetric/Signature/FALCON/1.2/falcon_512.cry
+++ b/Primitive/Asymmetric/Signature/FALCON/1.2/falcon_512.cry
@@ -2,10 +2,7 @@
  * @copyright Galois, Inc
  * @author Marios Georgiou <marios@galois.com>
  */
-module falcon_512 where
-
-submodule falcon_inst = spec2
-    where
+module falcon_512 = spec2 where
     type _k = 9
     type _n = 512
     type pkbytelen = 897
@@ -532,5 +529,3 @@ submodule falcon_inst = spec2
         (0.9996988186962042, -0.02454122852291245),
         (0.9999247018391445, -0.012271538285720572)
     ]
-
-import submodule falcon_inst


marsella (approved):
Looks good to me.

The other place I see this pattern is [in `falcon.tex`](https://github.com/GaloisInc/cryptol-specs/blob/master/Primitive/Asymmetric/Signature/FALCON/1.2/falcon.tex#L56)
```
\begin{code}
  module falcon where

  import falcon_512
\end{code}
```

Do you anticipate that being a problem for your use case? I'm guessing not, since it's easier to use the parameter-specific modules directly, but wanted to flag.

RyanGlScott:
I'm not sure what to make of `falcon.tex` as a piece of Cryptol code. Since it doesn't export anything, it's really only useful if you load the module into the Cryptol REPL and interact with it there.

That being said, I get that sense that the code is trying to accompany an existing LaTeX specification, and as such, there is a need to put _something_ in `falcon.tex`, since it's the "top-level" entrypoint to FALCON. (`sphincs.tex` [does something similar](https://github.com/GaloisInc/cryptol-specs/blob/9bb642049b3cb9175a56e91c839138b963d2c9b9/Primitive/Asymmetric/Signature/SphincsPlus/3.1/sphincs.tex#L65-L67).) I don't have any better ideas about what to put in `falcon.tex`, so I guess this is fine for now. (If Cryptol offered some some of re-export mechanism, we might consider re-exporting `falcon_512.tex`'s definitions from `falcon.tex`, but I don't believe this is currently possible in Cryptol.)

PR #362, opened 2026-09-08
clarify assumptions about `cswap` input

### LINKED ISSUE #361: `Curve25519`: `cswapTrue` property is falsifiable
`Common::EC::Curve25519`'s [`cswapTrue`](https://github.com/GaloisInc/cryptol-specs/blob/0de36bc45ab3e5cce9dc12fec41a88558377383c/Common/EC/Curve25519.cry#L342-L349) property is not true for all inputs, as seen below:

```
Common::EC::Curve25519> :prove cswapTrue 2 (1, -1)
Counterexample
cswapTrue 2 (1, -1) = False
(Total Elapsed Time: 0.016s, using "Z3")
```

I'm fairly certain that the `cswap` function is intended to only ever be called with `0` or `1` as its first argument, however, and that appears to be the case in practice. If so, it may be worth constraining `cswapTrue` to look more like:

```
property cswapTrue x = cswap 1 (cswap 1 x) == x
```

Alternatively, we might also consider making the `cswap` function take a `Bit` as its first argument instead of `[255]`, but it's not clear to me what the knock-on consequences of this change would be.

marsella:
Closes #361

### DIFF
--- a/Common/EC/Curve25519.cry
+++ b/Common/EC/Curve25519.cry
@@ -301,6 +301,10 @@ X25519 kb ub = result
  * "Conditional swap".
  * This function must be constant time in a production implementation.
  * Cryptol cannot enforce this.
+ *
+ * The procedure that uses `cswap` implicitly enforces that `swap` is
+ * always binary -- that is, it's a 255-bit string for which only the
+ * least significant bit is non-zero.
  */
 cswap : [255] -> (Z p, Z p) -> (Z p, Z p)
 cswap swap (x, y) = (BVtoZ `{p} (dummy ^ x_2), BVtoZ `{p} (dummy ^ x_3))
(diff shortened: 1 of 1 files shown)


Common/EC/Curve25519.cry
```diff
@@ -323,12 +327,12 @@ cswap' : [255] -> (Z p, Z p) -> (Z p, Z p)
...
+ * The swap implementations are equivalent for appropriate values of `swap`.
  * ```
  * :check cswapEq
  * ```
  */
-cswapEq s x = elem s [0, 1] ==> cswap s x == cswap' s x
+cswapEq x = (cswap 0 x == cswap' 0 x) && (cswap 1 x == cswap' 1 x)
 
 /**
  * `cswap False` is the identity.
```
RyanGlScott: `cswap False` doesn't quite typecheck, so perhaps this should be `cswap 0` instead.

Common/EC/Curve25519.cry
RyanGlScott: Similarly, perhaps this should be `cswap 1` instead.

Common/EC/Curve25519.cry
```diff
@@ -346,7 +350,7 @@ property cswapFalse x = cswap 0 x == x
  * :check cswapTrue
```
RyanGlScott: FWIW, you can actually `:prove` this in a relatively short amount of time if you pick a What4-based prover (e.g., `:set prover=w4-z3`). Similarly for `cswapFalse` and `cswapEq`.

RyanGlScott (approved):

Common/EC/Curve25519.cry
marsella: Oh great, thanks!

PR #360, opened 2026-09-05
SHA3: Explicitly name the underlying Keccak instance

glguy:
This enables external verification in to access the underlying Keccak primitives as used in a SHA3 instantiation. In particular this support compositional verification activities against refined specifications.

PR #365, opened 2026-09-11
AES-GCM-SIV: explicitly declare the AES instance

marsella:
This one did previously name the import, but it's nice to have it explicitly as a submodule, to match the pattern in #360 and #364.


glguy (approved):
Making the named submodule helps with SAW imports. Thanks!

